PR #751, opened 2022-03-02
Assign cells (WIP NOT FINISHED)

maxgestic:


minipunch:
coords prob need adjusting if #1043 is merged fyi

PR #823, opened 2022-04-17
Added c3swathawk to compatibleHeli list

DatBoiWeep:
pls approve mini 👼


minipunch:
Hello mister Christopher, thanks for the PULL REQUEST

However am I am going D E N Y this PR due to the fact that this change is actually not going to do anything since the `usa_policeheli` resource is the OLD heli cam resource that is not being used at the moment (notice it is not started in our `server.cfg`)

We now are using `ryans1230helicam`

So the file you want to update is here: https://github.com/minipunch/usarealismrp-fxserver/blob/2dac3acbfe14da5725f6ecf09edaa97a48054589/resources/ryans1230helicam/config.json

minipunch (changes requested):
see comment I left

minipunch (approved):

PR #826, opened 2022-04-18
Weapon dmg modifier

DatBoiWeep:
this was made by chango, but he too lazy to actually learn how to PR


minipunch:
Looks good, we should also try adding melee weapons to see if we can reduce the damage they do perhaps

minipunch:
Going to just deny this PR and add them manually as I made some changes and there is now a merge conflict that is non-trivial to resolve using the editor online

Committed this to public-server just now

DatBoiWeep:
Sounds good, Ill have Chang or myself look into the melee dmg

PR #828, opened 2022-04-19
Added assistcomm rank for sasp

DatBoiWeep:
let me know if I borked anything up sir


minipunch (changes requested):
Get rid of your changes in the `.gitignore` file as well as the `server.cfg` file and then I think this will be okay to merge

DatBoiWeep:
> Get rid of your changes in the `.gitignore` file as well as the `server.cfg` file and then I think this will be okay to merge

I got rid of the `server.cfg` and `.gitignore` file from the PR :)

DatBoiWeep (commented):
perfection 🔥

minipunch (approved):

PR #830, opened 2022-04-21
Update for new cars

Urbaniized:


minipunch:
I'll merge this

But note there is also the vehicle menu at the prison; that's under ``usa_prison``, so you probs want to add it there if it's intended for BCSO to use also @Urbaniized

minipunch (approved):

PR #832, opened 2022-04-24
New vehicles | Trailer

Urbaniized:


server.cfg
```diff
@@ -180,6 +179,8 @@ ensure 1979_Hoonigan_K5
 ensure 2021_Bronco
 ensure valor21durango
 ensure valor21tahoe
+ensure [GODz]
```
minipunch, line 182: You sure this works? I think you have to start each one within that folder individually if I am not mistaken

minipunch (approved):

PR #841, opened 2022-05-03
Toggleclothing update

DatBoiWeep:
Should be good for both mp male and female ped models :)


minipunch:
No need to copy the events and rename them like you did though

Just use the existing component events @DatBoiWeep

minipunch (changes requested):
see the comment I left

DatBoiWeep:
> No need to copy and paste the events and rename them like you did
> 
> 
> 
> Just use the existing component events but with the new numbers @DatBoiWeep 

Oooooh okay. Bet. I'll do that and also add the /bag command too while I'm at

DatBoiWeep:
Should be good now :)

minipunch (changes requested):
`/mask` seems broken :bruh:

DatBoiWeep:
I fixed it. Pls review /e peeposhy

resources/usa_toggleclothes/cl_toggleclothing.lua
```diff
@@ -58,28 +78,139 @@ AddEventHandler("headprops:toggleComponent", function(component_index)
...
+				else 
+					SetPedComponentVariation(ped, component_index, components[strComponentIndex].value, components[strComponentIndex].texture, 2)
+				end
+			end
+		end
+		-- End of pants
+	elseif component_index == 5 then -- /backpack
+		TriggerEvent("usa:playAnimation", "clothingshirt", "try_shirt_neutral_d", -8, 1, -1, 53, 0, 0, 0, 0,  1.5)
+		if GetPedDrawableVariation(ped, component_index) ~= 0 then -- component is on, take off
+			local value = GetPedDrawableVariation(ped, component_index)
```
minipunch, line 143: Would be ideal to not have so much code being copy and pasted around for cleanliness purposes

But I'll merge this for now since it works and perhaps in the future some of the duplicated code can be put into a shared function

minipunch (approved):

PR #844, opened 2022-05-04
Few Slight Changes

Savage912-Official:
**- Drug Price Change** - Raised the weed sell price and lowered the cocaine sell price on the streets due to having the ability to sell cocaine with deliveries, and weed takes longer to produce meaning it should have a higher reward.

**- Repair Kit Change** - Changed the repair kits sold at the hardware store to only be 8 pounds because there are other repair kits roaming around the city that only weight 8 meaning there is no need to have one that weighs 20 and one that weighs 8.

**- Delivery Truck Spawn** - Made the delivery truck spawn with a full tank of fuel rather than just spawning with a randomized tank of fuel each time a player would get one.


andrewclinger:
**Changed license weight** - Changed license weight due to you can not store or drop them when not using them.

Savage912-Official:
Yeah the issue is we had to upload the whole file due to github desktop not allowing us to clone the repo so we just uploaded the file from desktop.

minipunch:
> Yeah the issue is we had to upload the whole file due to github desktop not allowing us to clone the repo so we just uploaded the file from desktop.

See if you can do it another way so it doesn't have all those whitespace changes

Luckily GH has a feature to "hide whitespace changes" when viewing, but I don't want it to have weird formatting when I pull this change lol, I'll have to double check how it looks in a little, but if you can fix it beforehand that'd be ideal

Savage912-Official:
> > Yeah the issue is we had to upload the whole file due to github desktop not allowing us to clone the repo so we just uploaded the file from desktop.
> 
> See if you can do it another way so it doesn't have all those whitespace changes
> 
> Luckily GH has a feature to "hide whitespace changes" when viewing, but I don't want it to have weird formatting when I pull this change lol, I'll have to double check how it looks in a little, but if you can fix it beforehand that'd be ideal

Yep I fixed it just had to change the whitespace setting

minipunch:
> > > Yeah the issue is we had to upload the whole file due to github desktop not allowing us to clone the repo so we just uploaded the file from desktop.
> > 
> > 
> > See if you can do it another way so it doesn't have all those whitespace changes
> > Luckily GH has a feature to "hide whitespace changes" when viewing, but I don't want it to have weird formatting when I pull this change lol, I'll have to double check how it looks in a little, but if you can fix it beforehand that'd be ideal
> 
> Yep I fixed it just had to change the whitespace setting

Think that means it's just fixed for when you view it lol

But the problem is still there if you don't select 'hide whitespace changes'

minipunch:
I would recommend opening it up in git and viewing the diff on your local machine before pushing and make sure the diff only shows what u changed and not changes in the whole file

PR #846, opened 2022-05-04
feat(mechanicjob): Changed mechanics pay

Crownapple:
I believe that Mechanics mostly do their work in their own vehicle to drive around to help the citizens out. So I believe that the pay should be increased higher for fuel and other things.


resources/usa_mechanicjob/mechanic_sv.lua
```diff
@@ -2,7 +2,7 @@ exports["globals"]:PerformDBCheck("usa_mechanicjob", "mechanicjob")
 
 local NEEDS_TO_BE_CLOCKED_IN = true
 
-local TOW_REWARD = {100, 450}
+local TOW_REWARD = {100, 600}
```
minipunch, line 5: FYI this isn't the periodic paycheck amount

This is the reward range for when you tow a vehicle to the lot and impound it

See `usa_paycheck` for the periodic paycheck while clocked on, which I believe is what you intended to change here

minipunch (changes requested):

Issue #842, opened 2022-05-04
fuel loss when parking


importedskills:
When refueling vehicle and storing into house, fuel evaporates,
When refueling and storing into city garage, fuel stays.

Savage912-Official:
> When refueling vehicle and storing into house, fuel evaporates, When refueling and storing into city garage, fuel stays.

Bug has been fixed.

PR #849, opened 2022-05-04
feat(Properties): Vehicle Parking With Fuel

Savage912-Official:
Fixed vehicle fuel not saving when storing it into an owned property.

Issue found by: @importedskills
Issue Link: https://github.com/minipunch/usarealismrp-fxserver/issues/842


minipunch (approved):
Nice!

minipunch:
Although for next time:

I think this would be better labeled as a 'fix' instead of a 'feat' here, but nice job otherwise 👍

PR #850, opened 2022-05-04
Adjustment(Weed): Changed Weed Weights

Crownapple:
**Changed Weed Weights** - Changed the weights of bud and packaged weed because I believe packaged weed should weigh more than weed bud as packaged would have a bit more weed in it.


minipunch (commented):
What about the STEM that holds the raw weed buds though lol

Surely that would add weight, and then removing those stems to 'baggy' it would reduce weight right

Cause when harvesting, you don't just get a single 'bud'; you get a 'branch' with a stem that is full of multiple weed 'buds', even though the item name is "weed bud", maybe "weed branch" would be better LOL

PR #851, opened 2022-05-04
feat(Mechanic Job): Adjusted Mechanics Pay

Crownapple:
**Adjusted Mechanics Pay** - Believe that Mechanics mostly do their work in their own vehicle to drive around to help the citizens out. So I believe that the pay should be increased higher for fuel and other things.


minipunch (approved):

minipunch:
For future reference:

I wouldn't really consider this a 'feat', more so an 'adjustment' probably than feature really

PR #853, opened 2022-05-05
feat(Weed): Changed Weed selling price

Crownapple:
- **Drug Price Change** - Raised the weed sell price and lowered the cocaine sell price on the streets due to having the ability to sell cocaine with deliveries, and weed takes longer to produce meaning it should have a higher reward.


minipunch:
People also argue that because cocaine might be more 'dangerous' in a legal aspect that it would be worth MORE than weed

Though weed does require more time to produce

So it's just a matter of finding a balance, I think maybe leaving the coke prices or maybe lowering them a bit less than you have already done is fair

minipunch (changes requested):
see comment

PR #856, opened 2022-05-05
Spray paint reduced time

Khajit71:
- Reduced the time a graffiti lasts from 30 days to 14 days, as there are a lot that stay on for a long time, and affects people's resources, sometimes causing lag


minipunch (approved):
There's already optimizations in place that prevents displaying more than a set number of sprays at any given time to help with that

The only thing is maybe checking distance to sprays every frame but I don't have any evidence that it's really bad at all

But I'll still merge this for now

PR #863, opened 2022-05-08
Disable Radio Dispatch

Urbaniized:


minipunch:
We already have this code running in `usa_rp2`

PR #864, opened 2022-05-09
Reduced weed packaging time

Khajit71:
Reduced the time it takes to package weed, due to it only being a put the weed in the bag thing, and should take long. Reduced from 38 seconds to 10 seconds. Also due to weed coming in such high quantities(roughly 13 per plant)

### DIFF
--- a/resources/usa_weed/cl_weed.lua
+++ b/resources/usa_weed/cl_weed.lua
@@ -1,4 +1,4 @@
-local KEY = 38 -- "E"
+local KEY = 10-- "E"
 local SOUND_ENABLE = true
 local processed = false
 local harvested = false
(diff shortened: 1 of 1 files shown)


Savage912-Official:
Do you mean to change the button from E to PAGEUP?

resources/usa_weed/cl_weed.lua
```diff
@@ -167,7 +167,7 @@ function DrawText3D(x, y, z, distance, text)
         SetTextCentre(1)
         AddTextComponentString(text)
         DrawText(_x,_y)
-        local factor = (string.len(text)) / 380
+        local factor = (string.len(text)) / 100
```
minipunch, line 170: Lmao I have a feeling you didn't test this

Pretty sure this has nothing to do with the time it takes to package weed

minipunch:
1) see my comment

2) you're doing more than just putting the weed from the tree into a bag though, this processing stage involves trimming and what not also

PR #875, opened 2022-05-10
Adjustment(House): Number of owned houses

Crownapple:
I think you should be able to own more then 5 house as I have seen some suggest for it to be set higher.


minipunch:
Not sure we want to do this either

Some people say that people own too many houses and it's hard to find available properties for new people

Crownapple:
> Not sure we want to do this either
> 
> Some people say that people own too many houses and it's hard to find available properties for new people

TBH I've seen a lot of suggestions to have no limits two properties owned as residential but that also would not be fair to new people in the city because they would not have a chance to own a house that's why I think it should be capped at how many houses you can own but also I feel like people have different rp on their characters each and every other day and I'm like to have multiple houses different areas for rp

Crownapple:
> Not sure we want to do this either
> 
> Some people say that people own too many houses and it's hard to find available properties for new people

I think what is hard too find is businesses I will be adding a cap to them soon

PR #876, opened 2022-05-10
Adjustment(Cocaine): Changed Cocaine selling price

Crownapple:
I think Cocaine selling price should be changed due it being more dangerous and also has a % chance for the police to be notified when delivering.


minipunch:
Also not sure this needs to be raised so much

PR #877, opened 2022-05-10
Crownapple hangout mlo

Crownapple:
https://www.gta5-mods.com/maps/mlo-druglab


minipunch:
This spot is like the LAST place I think a 'hidden drug lab' would be 😂

minipunch (approved):

PR #880, opened 2022-05-11
Adjustment(Garage): Adjusted Mirror Park Garage

Savage912-Official:
Changed the mirror park garage location to a better spot right beside the old one so vehicles don't glitch when pulling out of garage.


resources/usa_garage/client.lua
```diff
@@ -58,13 +58,12 @@ locations = {
 	{x = 419.80194091797, y = -1629.1114501953, z = 29.282044067383, noBlip = true}, -- Davis, LS tow yard
 	{x = 4527.1513671875, y = -4529.048828125, z = 4.2344647407532}, -- cayo perico island
 	{x = 5171.826171875, y = -4635.3139648438, z = 2.5622141551971, noBlip = true},
-	{x=1121.6781, y=-774.7593, z=57.8709, noBlip = true}, -- mirror park repair shop
+	{x = 1110.33, y = -780.18, z = 58.34, noBlip = true}, -- mirror park repair shop
```
minipunch, line 61: Z value changed from `57` -> `58`

Did you make sure the circle wasn't floating?

Savage912-Official: Yep it doesn't float the spot a bit over I moved it to is a little higher ground that's why the Z value is changed.

minipunch (approved):

PR #883, opened 2022-05-12
Adjustment(LSCustoms): Config Change

Savage912-Official:
Changed the config to enable it so people cannot enter the bennys if a person is already in it to prevent glitching and not being able to see certain parts of your vehicle.


resources/lscustoms/cl_config.lua
```diff
@@ -911,10 +911,10 @@ LSC_Config.ModelBlacklist = {
 }
 
 --Sets if garage will be locked if someone is inside it already
-LSC_Config.lock = false
+LSC_Config.lock = true
 
 --Enable/disable old entering way
-LSC_Config.oldenter = false
+LSC_Config.oldenter = true
```
minipunch, line 917: What does this change?

Also you tested this right?

Savage912-Official: It just allows you to click enter instead of just auto putting you into the ls customs menu. Main change was people not being able to enter the customs while someone else was in it, but if needed this can be changed back.

All has been tested.

minipunch: I think people would rather not have to press anything to enter right?

I think auto pulling them in + not entering if someone is in there (like some text that says "please wait") would be best instead of forcing everyone to now press enter every time

Have you tried it that way?

Savage912-Official: That would be fine. We can just remove the oldenter and set it back to false.

Savage912-Official: Done

minipunch: Ok

I'm not familiar with how the script works with those values changed

So hopefully it all works smooth that way lol

minipunch (approved):

PR #887, opened 2022-05-12
Adjustment(Insurance): Changed price of Insurance

Crownapple:
I think the price of the Insurance should be changed to make them more realistic prices.


minipunch:
Not entirely sure I agree that they should be lowered

PR #902, opened 2022-05-15
Adjustment(CFG): Added Pizza Shop MLO

Savage912-Official:
Added Pizza Shop MLO to CFG as requested in a suggestion.


ChangyBoi:
I know a lot of people wouldnt be happy with this because of how close to legion it is. We already have 2 massive MLO such as pillbox and MRPD. If we can find a pizza MLO that isn't near legion I'm all for it but legion just needs to be kept as clear as possible IMO

Savage912-Official:
> I know a lot of people wouldnt be happy with this because of how close to legion it is. And could potentially worsen performance in that area, We already have 2 massive MLO such as pillbox and MRPD. If we can find a pizza MLO that isn't near legion I'm all for it but legion just needs to be kept as clear as possible IMO

Yeah I told mini we could get gabz which is on popular street but he said to just do this one for now so I did. I do think we should change to a different one at some point though.

minipunch:
Let's just go ahead and get Gabz' and we can add it instead

Savage912-Official:
> Let's just go ahead and get Gabz' and we can add it instead

Alright only issue is it’s through tebex so the FiveM account associated with the server license key has to buy it from him or it won’t allow us to use it.

minipunch:
outside textures seem to be flickering badly and makes it look bad

even 'cleared server cache'

PR #904, opened 2022-05-16
Adjustment(Death): Disable talking on radio while dead

Savage912-Official:
Disables talking on the radio while dead as I have seen multiple players doing this after being shot down.


ChangyBoi:
A M E N

minipunch (approved):

PR #910, opened 2022-05-16
Change(Guns): Weapon Changes

Savage912-Official:
- Removed Screen Shake When Shooting
- Lowered Recoil When Shooting

- Added New Guns
- Weapons Now Have Bullet Spread When Shooting Rather Than Recoil (Still a bit of recoil)

New Guns:
- AK-47
- Beretta M9A3
- Desert Eagle
- Glock
- Glock 18C (AP Pistol)
- M4 Rifle
- Sig Sauer P226R
- Remington 870
- Uzi

I do plan to add more weapons in the future just wanted to see how these do for now to see how everything goes with them in the server.


Crownapple:
+1 ME LIKE

resources/usa_ammunition/server/sv_ammunition.lua
```diff
@@ -72,7 +72,7 @@ local WEPS_NO_MAGS = {
         AMMO_NAME = "12 Gauge Shells",
         MAX_CAPACITY = 8
     },
-    [GetHashKey("WEAPON_PUMPSHOTGUN_MK2")] = {
+    [[--[GetHashKey("WEAPON_PUMPSHOTGUN_MK2")] = {
```
minipunch, line 75:
```suggestion
    [[--[[GetHashKey("WEAPON_PUMPSHOTGUN_MK2")] = {
```
isn't it supposed to be like that instead for a comment

minipunch:
1) > "Removed Some Old Weapons"

What were the 'old weapons' ?

and 2) does the 'bullet spread' still work when shooting from within a vehicle also?

resources/usa_ammunition/server/sv_ammunition.lua
Savage912-Official: Yeah it looks weird on github but its because I commented out multiple lines. 

Also when I say "Old Weapons" I mean I just commented out the weapons that weren't replaced with custom models until I could get them replaced with custom models and metas.

Savage912-Official:
> 2\. f

Regarding the bullet spread when shooting it should and the _weaponrealism script also makes weapons a bit more inaccurate while shooting from vehicles. I still do have a bit of recoil on the weapons as well.

resources/usa_ammunition/server/sv_ammunition.lua
Savage912-Official: Also no the way I commented it out was correct.

minipunch: > "I just commented out the weapons that weren't replaced with custom models until I could get them replaced with custom models and metas."

And why not just leave them in ?

resources/_weaponrealism/cl_weapon-realism.lua
```diff
@@ -404,68 +404,69 @@ end)
 
 -- custom (more noticable) weapon recoil --
 local RECOILS = {
```
minipunch, line 406: If our intention is to disable the recoil completely here; instead of replacing all of the existing values with 0 can you just make some constant that will act as a toggle variable to turn it on and off. That way we can easily go back without having to search through git history if we wanted.

resources/usa_ammunition/server/sv_ammunition.lua
Savage912-Official: We can. I'll change quickly

minipunch (changes requested):
see last comment

resources/_weaponrealism/cl_weapon-realism.lua
Savage912-Official: Yeah I can look into that.

ChangyBoi:
in the gun-shop_sv.Lua, you re-enabled stun gun, marksman pistol, and revolver. Those are to be left commented out. Whatever is left in the server of those will be what is left

Savage912-Official:
> in the gun-shop_sv.Lua, you re-enabled stun gun, marksman pistol, and revolver. Those are to be left commented out. Whatever is left in the server of those will be what is left
> 
> Also your meta files might go and conflict with the ones in gunmetas. If needed you can remove everything in there but leave the weaponanimations.meta since it controls the rate of fire

Fixed

Savage912-Official:
> see last comment

Everything is sorted now

Savage912-Official:
Gonne redo later without custom models and just with bullet spread and lower recoil.

PR #912, opened 2022-05-17
New vehicle damage script

Urbaniized:


minipunch:
How is this any different?

Where did u get it?

PR #914, opened 2022-05-17
(Hunting): Raised prices for fur and meat

OffixialMikey:
I'm submitting this change since I believe that Hunting should have a little more reward for the hard work you have to do during this job.


minipunch (approved):
"HARD WORK"

RIP

PR #919, opened 2022-05-18
Fix(Radio): Fix Talking on Radio while Dead

Savage912-Official:
Should fix the issue with still being able to talk on the radio while downed.


resources/rp-radio/client.lua
```diff
@@ -526,7 +526,7 @@ Citizen.CreateThread(function()
         local broadcastType = 3 + (HAS_EARPIECE and 1 or 0) + (radioConfig.AllowRadioWhenClosed and 1 or 0) + ((Radio.Open and radioConfig.AllowRadioWhenClosed) and -1 or 0)
         local broadcastDictionary = Radio.Dictionary[broadcastType]
         local broadcastAnimation = Radio.Animation[broadcastType]
-        local isBroadcasting = IsControlPressed(0, radioConfig.Controls.Broadcast.Key) and not IsPlayerFreeAiming(PlayerId()) and GetLastInputMethod(0)
+        local isBroadcasting = IsControlPressed(0, radioConfig.Controls.Broadcast.Key) and not IsPlayerFreeAiming(PlayerId()) and GetLastInputMethod(0) and not IsEntityDead(GetPlayerPed(-1))
```
minipunch, line 529:
```suggestion
        local isBroadcasting = IsControlPressed(0, radioConfig.Controls.Broadcast.Key) and not IsPlayerFreeAiming(PlayerId()) and GetLastInputMethod(0) and not IsEntityDead(playerPed)
```

For optimization purposes, no need to duplicate native calls here since we already saved the ped into a variable just above

minipunch (approved):

PR #922, opened 2022-05-19
Adjustment(Weapons): Disable screen shake when shooting.

Savage912-Official:
This disables the screen shaking like crazy when shooting a weapon. it's not needed since we have recoil for weapons.


minipunch:
This was kinda a feature not a bug lol

Can u maybe show some before and after videos for people (myself included) to decide which they like better

PR #923, opened 2022-05-20
New(Crosshair): Added new crosshair

Savage912-Official:
Added a new crosshair and removed the old one.

Suggested By: Khajit


resources/usa-crosshair/client/cl_crosshair.lua
```diff
@@ -0,0 +1,22 @@
+local crosshairActive = false
+local disableCrosshair = false
+
+RegisterCommand("crosshair", function()
+    disableCrosshair = not disableCrosshair
```
minipunch, line 5: You sure this works with our chat command system?

Pretty sure this would only work in the F8 console

minipunch (changes requested):
Can you utilize the existing database/code to ensure their crosshair setting is persistent across multiple sessions please

Right now someone who wants it other than the default would have to use the command EVERY time they  play

Savage912-Official:
Yeah let me go look at the database stuff

PR #924, opened 2022-05-20
Adjustment (Drug Prices)

Khajit71:


Savage912-Official:
I would say the meth and lsd need a raise but not sure that the weed needs to be buffed due to how long it takes to grow it.

Crownapple:
I agree on the meth and lsd but do not agree on the weed changes

minipunch (approved):

PR #927, opened 2022-05-21
New panic button sound

ChangyBoi:
new panic button sound also decreased volume to hopefully prevent sound inducing migraine :) 

Sound used https://www.youtube.com/watch?v=STesSKOUyCY


minipunch:
FYI you can also adjust the volume in the `/p` command where it triggers the sound, it's just an argument to the event

minipunch (approved):

PR #929, opened 2022-05-22
Adjustment(Stores): Shovel & Pickaxe Weights

Savage912-Official:
Lowered the shovel and pickaxe weights from 30 to 15 as they should not be 30.


minipunch:
Don't really agree that they should be 15 units instead of 30

A shovel and pickaxe are pretty large and heavy IRL

Remeber, it's not "pounds", it's just how much "space" it might take for a person to realistically carry

PR #930, opened 2022-05-22
Adjustment(Mining): Adjusted Mining Radius

Savage912-Official:
Adjusted Mining Radius to be a bit smaller as it was to big of a radius


minipunch:
What is this? For drawing text and listening for the keypress I am guessing

minipunch (approved):

PR #947, opened 2022-05-29
Recoil update

ChangyBoi:
Removed camera shake - this only affected the people outside of cars - Made it unfair for people fighting on foot

Lowered the car inaccuracy pitch - this only affects 3rd person shooting especially from helicopters meaning you shoot once you would aim at the ground/sky

Overall everything got lowered a little and should make gunfights smoother

I might do dmg changes later down the road in case its needed


minipunch:
Assuming you tried the changes out on ur own server and they were good

minipunch (approved):

PR #952, opened 2022-06-03
Vehicle damage update

Urbaniized:


Urbaniized:
Reduced side and rear damage aka fuel tank damage

minipunch (approved):

PR #953, opened 2022-06-05
Character Creation: Removed BCSO Rank Check

maxgestic:
Now that DB is reset and no-one will have previous system characters, this is unnecessary.


minipunch (approved):

minipunch:
Nice 👍

PR #959, opened 2022-06-08
Hopefully fixed taser weight

DatBoiWeep:
I changed the original 0.2 value and put it into a format to maybe avoid it from having multiple decimals???

It works perfectly fine on my test server, but for some reason is broken on the live server, causing it to be 0.1999..7


minipunch:
I doubt this will fix that issue cause ultimately that expression evaluates to the same thing, lol: `0.2`

So it might be something happening later in the code that's causing it, but I need some repro steps

Can you define a set of steps to reproduce the problem so I can try to reproduce it

DatBoiWeep:
> I doubt this will fix that issue cause ultimately that expression evaluates to the same thing, lol: `0.2`
> 
> So it might be something happening later in the code that's causing it, but I need some repro steps
> 
> Can you define a set of steps to reproduce the problem so I can try to reproduce it

Pretty much, just buy a cartridge at the armory on the live server, and it's given a weight of 1.999997
But when testing it on the test server, it works perfectly fine

DatBoiWeep:
Im going to close this for now :)

PR #967, opened 2022-06-10
Removed Ford Hula Girl from shop

Urbaniized:


minipunch:
What's the issue with it lol

minipunch (approved):

Urbaniized:
It doesn't drive right, collisions hit the ground and it doesn't turn right. Haven't been able to figure out what the issue is

PR #968, opened 2022-06-10
Disable police scanner

Urbaniized:


resources/usa_rp2/client.lua
```diff
@@ -226,6 +226,8 @@ Citizen.CreateThread(function()
     end
 end)
 
+SetAudioFlag('PoliceScannerDisabled', true)
```
minipunch, line 229: Would be good to leave a comment on why this is necessary

I'll let it slide this time cause it's relatively self explanatory, but just for future reference

minipunch (approved):

PR #993, opened 2022-06-20
Pd changes

DatBoiWeep:
Removed the comments out since we don't need them anymore
and removed the bathroom locker room


maxgestic:
How comes MRPD locker for BCSO is getting removed

DatBoiWeep:
> How comes MRPD locker for BCSO is getting removed

SASP Command requested the removal in their recent meeting

minipunch (approved):

PR #998, opened 2022-06-22
Updated Warehouse main.lua

Khajit71:
Removed warehouse location `10` as it caused issued whenever entering/leaving. @PickleRick can also confirm, and it always happens to me too


minipunch:
For now I will merge this

But a better solution is to ensure "collisions" are loaded properly via the "request collisions" native and only teleporting when collisions to the warehouse are loaded

It happens because they get unloaded when the player is far away from the warehouse interior in the map (i.e. at the island when entering)

minipunch (approved):

PR #1059, opened 2022-07-10
Removed the drag and carry command while jailed

DatBoiWeep:


resources/usa_drag/client.lua
```diff
@@ -10,19 +10,21 @@ local isCarried = false
...
 	end
-	if not isDragging and not isDragged and not isCarrying and not isCarried then
+	if not isDragging and not isDragged and not isCarrying and not isCarried and not jailed then
 		TriggerEvent("usa:getClosestPlayer", 1.65, function(player)
     		if player then
 	        	local closestPed = GetPlayerPed(GetPlayerFromServerId(player.id))
 				if player.id ~= 0 and not IsPedInAnyVehicle(PlayerPedId()) and IsEntityVisible(closestPed) then
 					TriggerServerEvent('drag:sendDragPlayer', player.id)
-					--print('sent??')
+					print('sent??')
```
minipunch: plz remove debug prints when pushing live

resources/usa_drag/client.lua
```diff
@@ -135,11 +137,16 @@ AddEventHandler('drag:toggleCarryAction', function(toggleOn, _source)
 	ClearPedTasksImmediately(PlayerPedId())
 end)
 
+RegisterNetEvent("death:toggleJailed")
+AddEventHandler("death:toggleJailed", function(toggle)
```
minipunch, line 141: If we're going to do it this way, can you at least rename the event from `death:toggleJailed` to something more generic so it makes more sense when used not in the `usa_death` script

Something like `usa:toggleJailedStatus`

minipunch (changes requested):
see comments

resources/usa_drag/client.lua
DatBoiWeep: Removed :)

resources/usa_drag/client.lua
DatBoiWeep: On it boss

minipunch (approved):

minipunch:
hoping you ctrl + shift + f'd the whole repo to find all the scripts to rename all the events lul

PR #1060, opened 2022-07-11
Changed LS spawn

DatBoiWeep:
Spawn for LS will now put you in the Mall MLO where you can get just about everything you need. Plus a garage and bike shop are near it :)


minipunch:
Is there already a garage there? I don't see that you added one, just a bike shop @DatBoiWeep

minipunch:
Ok

We'll also need to update the spawn zone safe area for LS in the rules

minipunch (approved):

PR #1061, opened 2022-07-11
Changed wakeup location while jailed

DatBoiWeep:
They shall now wake up in the cell block


DatBoiWeep:
image
This is my test subject

minipunch (approved):

PR #1063, opened 2022-07-11
Blackmarket rotations

ChangyBoi:
Moved blackmarket to new positions

lowered the distance you see the prompt at


DatBoiWeep:
Cute

PR #1084, opened 2022-07-18
More staff logs

DatBoiWeep:
- Added a bunch of logs for CA+ to have in the event our Lord mini is not available


resources/swayam_essentials/server.lua
```diff
@@ -72,6 +72,8 @@ end, {
 TriggerEvent('es:addGroupCommand', 'gotowp', 'mod', function(source, args, char)
 	TriggerEvent("usa:notifyStaff", '^2^*[STAFF]^r^0 Player ^2'..GetPlayerName(source)..' ['..source..'] ^0 has teleported to waypoint.')
 	TriggerClientEvent("swayam:gotoWP", source)
+	Wait(1000) -- Need this in order to register new Coords
+	TriggerEvent("chat:sendToLogFile", source, "used /gotowp and went to the following coords ["..GetEntityCoords(GetPlayerPed(source)).."]")
```
minipunch, line 76: Did you test this? Because when you call `Wait()` on the server then usually the `source` variable gets lost

So I recommend saving the source into a new variable, call it `src` or smth and use it instead

minipunch (approved):

resources/swayam_essentials/server.lua
DatBoiWeep: I did test it and it worked out, but I can add it to a variable if needed

DatBoiWeep: Yeah, gave it a test again with me and a test dummy (ex-mod) and the logs showed the coords where we started noclipping and ended noclipping at :)

minipunch: Oh I see why

It works fine because `source` is defined as a local variable since it's a parameter to the `addGroupCommand` callback

PR #1099, opened 2022-07-22
Change(Police): Updated Prop for Police Badge Animation

Savage912-Official:
Updated the prop for the police badge animation and removed the old prop.


resources/dpemotes/__resource.lua
```diff
@@ -19,6 +19,4 @@ server_scripts {
 	'@mysql-async/lib/MySQL.lua',
 	'Server/Server.lua',
 	'Server/Updates.lua'
-}
-
-data_file "DLC_ITYP_REQUEST" "copbadge.ytyp"
```
minipunch, line 24: Wonder why the old badge has this .ytyp file but the new doesn't and if it really matters at all

minipunch (approved):

PR #1140, opened 2022-08-15
GSR and Evidence drop fix

DatBoiWeep:


resources/usa_evidence/cl_evidence.lua
```diff
@@ -154,7 +154,7 @@ Citizen.CreateThread(function()
 	while true do
 		Citizen.Wait(10)
 		local playerPed = PlayerPedId()
-		if IsPedShooting(playerPed) and WeaponDropsEvidence(GetSelectedPedWeapon(playerPed))  then
+		if IsPedShooting(playerPed) and not WeaponDropsEvidence(GetSelectedPedWeapon(playerPed)) then
```
minipunch, line 157:
```suggestion
		if IsPedShooting(playerPed) and not NO_EVIDENCE_WEPS[GetSelectedPedWeapon(playerPed)] then
```

minipunch: and then you can remove the `WeaponDropsEvidence` function all together (but make sure it's also not being used anywhere else first)

minipunch (changes requested):

resources/usa_evidence/cl_evidence.lua
DatBoiWeep: I looked into (Ctrl+Shift+F) the `WeaponDropsEvidence` seems like it's used nowhere else other than here. So she's mint

minipunch (approved):

minipunch:
👍 nice

PR #1166, opened 2022-08-23
Aim assist removal when using controller

DatBoiWeep:
Basically what the title says :)


minipunch (changes requested):
If you read the forum post comments people say this does more than just stop controller auto aim

It also prevents locking on when using any melee weapon for example (like squaring up to people with fists or melee weapons) which is not really something we want to do

And a couple folks also said it effects the `IsPlayerFreeAiming` native which we make use of in several scripts so whatever script uses it we want to test and make sure is not broken also

So let's either try 1) maybe blacklisting this for code for ALL melee weapons (not just fists as you have it now) and see if that is good enough

or 2) another option that maybe you want to try first could be using this code/native someone recommended instead:

```Citizen.CreateThread(function()
  while true do
    Citizen.Wait(0)
    SetPlayerTargetingMode(2)
  end
end)```

**Comment by DatBoiWeep** [comment]
> If you read the forum post comments people say this does more than just stop controller auto aim
> 
> 
> 
> It also prevents locking on when using any melee weapon for example (like squaring up to people with melee weapons) which is not really something we want to do
> 
> 
> 
> And a couple folks also said it effects the `IsPlayerFreeAiming` native which we make use of in several scripts so whatever script uses it we want to test and make sure is not broken also
> 
> 
> 
> So let's either try 1) maybe blacklisting this for code for ALL melee weapons (not just fists as you have it now) and see if that is good enough
> 
> 
> 
> or 2) another option that maybe you want to try first could be using this code/native someone recommended instead:
> 
> 
> 
> ```
> 
> Citizen.CreateThread(function()
> 
>   while true do
> 
>     Citizen.Wait(0)
> 
>     SetPlayerTargetingMode(3)
> 
>   end
> 
> end)
> 
> ```
> 
> 
> 
> 

Hmm okie. I can work on adding all our melee weapons, and if it doesn't work out. I'll try option 2 😃

**Comment by DatBoiWeep** [comment]
The 2nd option you provided worked better than the first option lol

File: resources/interaction-menu/cl_interaction.lua [hunk 1]
```diff
@@ -1306,6 +1306,14 @@ Citizen.CreateThread(function()
 	end
 end)
 
+-- disable auto lock when using controllers
+Citizen.CreateThread(function()
+	while true do
+	  Citizen.Wait(0)
```
minipunch, line 1312:
```suggestion
	  Citizen.Wait(100)
```
for performance reasons, u can check `F8 > resmon` to see how necessary it is (if the number goes up when set to wait 0); but for now, to be safe, let's bump this up to every 100 frames instead of every frame just in case

And u can/should test to confirm it still works with this change, even tho I believe it will

minipunch (approved):

PR #1169, opened 2022-08-25
Meth bug fix

DatBoiWeep:
- When completing the cooking process for meth, it will now be dropped on the floor if the player's inventory is full.


resources/usa_meth/sv_meth.lua
```diff
@@ -18,11 +18,24 @@ AddEventHandler("methJob:methProduced", function(itemName, securityToken)
...
+	local newCoords = {
+		x = methProduced.coords.x,
+		y = methProduced.coords.y,
+		z = methProduced.coords.z
+	}
+	newCoords.x = newCoords.x + (math.random() * 0.5)
+	newCoords.y = newCoords.y + (math.random() * 0.5)
+	newCoords.z = newCoords.z - 0.85
+	methProduced.coords = newCoords
+	methProduced.coords = GetEntityCoords(GetPlayerPed(char.get("source")))
```
minipunch, line 31:
```suggestion
```

this overwrites the lines just before it, so it makes them pointless

so if u want the item to spawn in a random spot around the player and not just on the exact player coords every time, delete this line

and fyi it's the same exact thing as line 21, just called a bit differently

minipunch (changes requested):

resources/usa_meth/sv_meth.lua
DatBoiWeep: oh LOL

DatBoiWeep: I will fix sir

DatBoiWeep: Removed it 😅

minipunch (approved):

PR #1172, opened 2022-08-27
feat(Police): Get Vehicle Records CMD

Savage912-Official:
Usage: /records ID

The Idea behind this was for cops to be able to pull vehicle records without having to contact a staff member to do it.

**Command is only temp as I will be adding this to the Person Search in the MDT.**

### DIFF
--- a/resources/usa_police/server.lua
+++ b/resources/usa_police/server.lua
@@ -261,7 +261,7 @@ end, {
 
 TriggerEvent('es:addJobCommand', 'records', {'sheriff', 'police' , 'judge', 'corrections'}, function(source, args, char)
 	if args[2] then
-		local user = exports["essentialmode"]:getPlayerFromId(tonumber(args[2]))
+		
 		local targetchar = exports["usa-characters"]:GetCharacter(tonumber(args[2]))
 		if targetchar then
 			local ownedVehicles = targetchar.get("vehicles")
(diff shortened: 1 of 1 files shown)


resources/usa_police/server.lua
```diff
@@ -259,6 +259,52 @@ end, {
     }
 })
 
+TriggerEvent('es:addJobCommand', 'records', {'sheriff', 'police' , 'judge', 'corrections'}, function(source, args, char)
+	if args[2] then
+		local user = exports["essentialmode"]:getPlayerFromId(tonumber(args[2]))
```
minipunch, line 264:
```suggestion
		
```

Line not needed?

minipunch (changes requested):

resources/usa_police/server.lua
```diff
@@ -259,6 +259,52 @@ end, {
...
+				if insurance.planName then
+					displayInsurance = "Valid"
+				end
+                TriggerClientEvent('chatMessage', source, "", {255, 255, 255}, " " .. targetchar.getFullName() .. "'s Citizen Records")
+                TriggerClientEvent('chatMessage', source, "", {255, 255, 255}, "^2Vehicles: ^0" .. vehiclenames .. " | ^2Insurance: ^0" .. displayInsurance .. " |")
+            end)
+		end
+	end
+end, {
+	help = "View citizen records"
```
minipunch, line 305: Would be good to get help text for the params also, here it would just be the ID param

minipunch (approved):

minipunch:
👍

PR #1176, opened 2022-08-30
update(Police): Added ID Params to Records CMD

Savage912-Official:
When typing /records it will now show the information needed for the command.


minipunch (approved):
Nice !

PR #1178, opened 2022-08-30
update(Police): Impound now sends a chat msg

Crownapple:
When impounding a vehicle it now sends a chat message


resources/usa_rp2/server.lua
```diff
@@ -236,7 +236,9 @@ AddEventHandler("usa:loadPlayerComponents", function(id)
 	TriggerClientEvent("usa:setPlayerComponents", usource, appearance)
 end)
 
-TriggerEvent('es:addJobCommand', 'impound', { "sheriff", "ems", "corrections" }, function(source, args, char)
+TriggerEvent('es:addJobCommand', 'impound', { "sheriff", "ems", "corrections" }, function(source, args, char, location)
+	local msg = "Person with SSN "..source.." calls for State Tow"
+	exports["globals"]:sendLocalActionMessageChat(msg, location, 15)
```
minipunch, line 241: Actually instead of a chat message we should simulate as if they did `/me` and do the 3d text on the ped

There are examples somewhere, just look for the existing `/me` command code

minipunch (changes requested):

resources/usa_rp2/server.lua
Crownapple: Ok

Issue #1177, opened 2022-08-30
make sure unbanning works with new admin menu


minipunch:
`erp_adminmenu`'s unbanning seems to not be working

we enter the ban ID but the ban record is not deleted automatically

- [x] make sure manually unbanning works
- [x] make sure automated temp ban unbanning works

minipunch:
@Savage912-Official ^^

If u can check that out that would be good

I can also look

minipunch:
💯 🥇

minipunch:
Did u also make sure the automated temp unbanning works too lol

Like by banning urself for a few seconds/minutes and rejoining and seeing that u actually do get unbanned in the right time @Savage912-Official

Savage912-Official:
> Did u also make sure the automated temp unbanning works too lol
> 
> Like by banning urself for a few seconds/minutes and rejoining and seeing that u actually do get unbanned in the right time @Savage912-Official

I did and it works as well

minipunch:
Ok nice

PR #1179, opened 2022-08-30
fix(AdminMenu): Unban Perms Fix

Savage912-Official:
It worked for me and @Crownapple because when I first tested it I did

local userGroup = exports["essentialmode"]:getPlayerFromId(source).getGroup()
if userGroup == "owner" or userGroup == "superadmin" or userGroup == "admin" and banid then

But I changed it after we tested to

local group = exports["essentialmode"]:getPlayerFromId(source).getGroup()

I just forgot to change

if userGroup == "owner" or userGroup == "superadmin" or userGroup == "admin" and banid then

to

if group == "owner" or group == "superadmin" or group == "admin" and banid then

so the perms weren't registering for unban

Issue #1177 can now be closed


minipunch:
Left another comment for you on #1177

minipunch (approved):

PR #1182, opened 2022-08-31
update(Impound): Impound CMD Updates

Savage912-Official:
The /impound command will no longer work if 3 or more mechanics are on duty. If there are not 3 or more mechanics on duty then the command will function as it used to.


minipunch:
For anyone wondering: this is to promote interaction between players versus have (usually LEOs) instantly delete a vehicle after every pursuit and what have you

minipunch (approved):

PR #1196, opened 2022-09-02
Prison contraband fixes

DatBoiWeep:
**Patch Notes**
- Fixed Shiv not working properly (My shitty spelling on quantity, broke it 😢)
- Also made some contraband items not stackable because that caused duplication issues. (Like I ended up with 8 shivs from only picking up 2 shivs LOL)

**Known Issue**
- Users with old Shiv can not drop/remove it from inventory.
   - Solution:  Have a cop just `/seize contraband` which will yeet it from their inventory.

**Notes for Mini**
- Not sure if you want to also have `Advanced Picks` not stackable, but thought it was better to not be stacked


minipunch:
Not stacking picks is prob fine

👍

minipunch (approved):

PR #1198, opened 2022-09-02
Potential Fix for Liveries/Neons/LowGrip

Urbaniized:
Will definitely need testing on live. 

It worked perfectly find on my server. Setting the `liveryindex` to 0 will actually apply the stock livery on the car and will not make it change every time it is pulled out of the garage

Also, for some reason, `liveryindex` did not affect the neons or low grips when it was set to 0


Urbaniized:
One thing I noticed, the DB got reset for the car I was testing when I set the `liveryindex` to 0, so it might have a downside of potentially removing all the current customizations on current cars.

Urbaniized:
Disregard last comment, after a couple more tests. The DB stays the same for other mods, only thing that changes is `liveryindex` goes from -1 to 0

minipunch:
R u saying u could reproduce losing low grips also ?

Urbaniized:
Couldn't consistently do it, only happened once. Def a DB value issue but doing this seemed to make everything work smoothly

minipunch:
Hm ok............................

I guess we can merge it and see how things go

minipunch:
Might be good if u can link this PR in any potentially related issues from the list

minipunch (approved):

Issue #1190, opened 2022-09-01
try fun looking orange picking script


minipunch:
https://forum.cfx.re/t/paid-standalone-esx-collecting-oranges-ui-realistic-collection-system/4904117

minipunch:
If ur looking at this and think u wanna try adding it 4 me

I will buy it 4 u just lmk

maxgestic:
But why though 🤔

minipunch:
Why NOT???

It's a way for civs to earn money lol

And I am sure some wouldn't mind / would like doing it

Crownapple:
Done

Savage912-Official:
Good to be closed

PR #1200, opened 2022-09-02
feat(Oranges): Orange Collecting with UI

Crownapple:
Today I present to you the collecting oranges, thanks to which you will be able to collect oranges on all orange trees on maps, and by the way there are 185 of them! It is possible to collect fruits on one tree only 1 time per restart for all players it is synchronized.

Multiple Sell / Buy Locations


minipunch (changes requested):
What's the deal with the two resources being added, one with `_lib` appended?

And 2 of the files I've looked at are basically identical but with some little changes?

Savage912-Official:
> What's the deal with the two resources being added, one with `_lib` appended?
> 
> And 2 of the files I've looked at are basically identical but with some little changes?

It's the way the script came when it was bought but
Burevestnik_orangejob - Handles the orange picking / orange picking UI
Burevestnik_orangejob_lib - Handles the orange factory / orange factory UI

minipunch:
Love the mini game when picking and that seems to work p good

But the orange selling at the factory is very confusing at first

Can u try to add better notifications for that somehow?

The issue is the notifications show *behind* the selling UI so u can't see em

minipunch:
Also would be cool if we could make the orange tree blips go away after someone picks them (but the other thing is more important)

minipunch (changes requested):
see comments

Savage912-Official:
> Love the mini game when picking and that seems to work p good
> 
> But the orange selling at the factory is very confusing at first
> 
> Can u try to add better notifications for that somehow?
> 
> The issue is the notifications show _behind_ the selling UI so u can't see em

Yeah i'll look into it. Also made the folder changed we talked about.

minipunch:
Niceeeeeeee @Savage912-Official 

Looks a lot better after being able to see the notifications now 👍 

Last thing before merging this is I think we should prob lower the payout just a bit

I got 360 for one tree, so like one litte section is 10k... and there's like 40-50k ish for all of them per restart for oranges

I guess it's not terrible

maybe I'll just lower a tad bit and push this 👍

minipunch:
Oh.. also one critical thing..

U gotta make sure player's can't use lua injectors (like eulen menu or what have u) to trigger the server events arbitrarily

U should

1) 'tokenize' the server event, see `salty_tokenizer` and how other scripts use the `securityToken` by sending it as an arg along with the triggered server event and it gets validated on the server sided script

2) add some server sided script logic/code that validates the event by making sure this player is actually supposed/allowed to be triggering it

Savage912-Official:
> Oh.. also one critical thing..
> 
> U gotta make sure player's can't use lua injectors (like eulen menu or what have u) to trigger the server events arbitrarily
> 
> U should
> 
> 1. 'tokenize' the server event, see `salty_tokenizer` and how other scripts use the `securityToken` by sending it as an arg along with the triggered server event and it gets validated on the server sided script
> 2. add some server sided script logic/code that validates the event by making sure this player is actually supposed/allowed to be triggering it

Sounds good! Will look into this in a bit

minipunch:
linking to #1190

resources/[orange-picking]/Burevestnik_orangejob_lib/server.lua
```diff
@@ -10,7 +10,10 @@ if Config.UseESX == true then
     local orange = { name = "Orange", type = "misc", price = 5, legality = "legal", quantity = 5, weight = 2 }
 
     RegisterServerEvent('bur_orangejob_lib:buy')
-    AddEventHandler('bur_orangejob_lib:buy', function(amount, price)
+    AddEventHandler('bur_orangejob_lib:buy', function(amount, price, securityToken)
```
Comment (minipunch) [inline resources/[orange-picking]/Burevestnik_orangejob_lib/server.lua:13]: Don't really need a security token for this function

Instead just make sure the price the client is sending is > 0, u can use `price = math.abs(price)` right at the start of the function

Otherwise the client could send a number like `-1000` for price and end up EARNING the money (since subtracting a negative number gives u addition)

minipunch (changes requested):
- [ ] make my requested changes for the buying function
- [ ] also have to edit the ``_resource.lua` or `fxmanifest.lua` to include the init script for the tokenizer in server and client
- [ ] this is good for doing 1) but we should also do 2) from earlier because u can bypass tokens if u know how and have the rite tools

resources/[orange-picking]/Burevestnik_orangejob_lib/server.lua
Comment (Savage912-Official) [reply resources/[orange-picking]/Burevestnik_orangejob_lib/server.lua]: Submitted all changes now just need to do "2) from earlier because u can bypass tokens if u know how and have the rite tools"

Comment (Savage912-Official) [reply resources/[orange-picking]/Burevestnik_orangejob_lib/server.lua]: Everything from the list is completed

resources/[orange-picking]/Burevestnik_orangejob/config.lua
```diff
@@ -15,4 +15,192 @@ Config.ItemOrangeDB = 'Orange'
 
 Config.Translation = {
 	['getorange'] = 'Collected oranges: ~g~',
+}
+
+Config.Tree_Locations = {
+	{coords = vector3(-1844.6024, 2008.6498, 130.91185)},
```
Comment (minipunch) [inline resources/[orange-picking]/Burevestnik_orangejob/config.lua:21]: I feel like these should be recorded somewhere already ? since it puts blips on the maps...

minipunch (approved):
Looks good otherwise, nice work

Issue #866, opened 2022-05-10
add /feedback [msg] command


minipunch:
so players can easily leave feedback that will be sent to a staff-only Discord channel via webhook

Savage912-Official:
Completed in #1214

minipunch:
added

PR #1214, opened 2022-09-06
feat(Commands): Feedback CMD

Savage912-Official:
Ever wanted to provide feedback about the server to the staff team? You can do so by typing /feedback message in chat!

You can only use the command once per server restart.
Usage: /feedback message

Issue: #866 Completed

### DIFF
--- a/resources/usa_rp2/server.lua
+++ b/resources/usa_rp2/server.lua
@@ -436,20 +436,21 @@ end, {help = "View the server's discord link."})
 ------------------------
 -- Feedback Command --
 ------------------------
-local cooldowns = {}
+local hasUsedCommand = {}
 TriggerEvent('es:addCommand','feedback', function(source, args, char, message)
+	local WEBHOOK_URL = GetConvar("feedback-webhook", "")
 	local char = exports["usa-characters"]:GetCharacter(source)
 	local char_name = char.getFullName()
 	table.remove(args, 1)
 	local message = table.concat(args, ' ')
 	local msg = "**Server ID:** "..source.."\n**Steam Name:** "..GetPlayerName(source).."\n**Steam ID:** "..GetPlayerIdentifiers(source)[1].."\n**Character Name:** "..char_name.."\n**Character Playtime:** "..FormatSeconds(char.get("ingameTime")).." \n**Feedback:** "..message.."\n"
-	
-	if not cooldowns[source] or ((os.clock() - cooldowns[source]) > 86400) then
-		cooldowns[source] = os.clock()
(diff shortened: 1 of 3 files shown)


resources/usa_rp2/server.lua
```diff
@@ -432,3 +432,50 @@ TriggerEvent('es:addCommand', 'discord', function(source, args, char)
...
+------------------------
+local cooldowns = {}
+TriggerEvent('es:addCommand','feedback', function(source, args, char, message)
+	local char = exports["usa-characters"]:GetCharacter(source)
+	local char_name = char.getFullName()
+	table.remove(args, 1)
+	local message = table.concat(args, ' ')
+	local msg = "**Server ID:** "..source.."\n**Steam Name:** "..GetPlayerName(source).."\n**Steam ID:** "..GetPlayerIdentifiers(source)[1].."\n**Character Name:** "..char_name.."\n**Character Playtime:** "..FormatSeconds(char.get("ingameTime")).." \n**Feedback:** "..message.."\n"
+	
+	if not cooldowns[source] or ((os.clock() - cooldowns[source]) > 86400) then
```
minipunch, line 447: Where does the `86400` number come from ?

resources/usa_rp2/server.lua
```diff
@@ -432,3 +432,50 @@ TriggerEvent('es:addCommand', 'discord', function(source, args, char)
...
+		cooldowns[source] = os.clock()
+		exports['erp_adminmenu']:sendToDiscord('New Feedback Submitted', msg, "16711680", GetConvar("feedback-webhook", ""))
```
minipunch, line 449: I think we have an export for sending discord messages in `resources/[system]/globals` server script

resources/usa_rp2/server.lua
minipunch: Also if we are doing once per restart then you don't need to worry about using `os.clock` at all right

You can just add them to the `cooldowns` table and check if they have already been added or not

And the restart will reset the table

minipunch (changes requested):

resources/usa_rp2/server.lua
Savage912-Official: > Where does the `86400` number come from ?

That was 24hr but working on updating it now

resources/usa_rp2/server.lua
Savage912-Official: > I think we have an export for sending discord messages in `resources/[system]/globals` server script

Working on this now

minipunch (approved):
Nice

PR #1218, opened 2022-09-06
feat(Wheels): Wheel Fall Off / Wheel Damage Script

Savage912-Official:
This script will fully overhaul the wheel damage system. Tires can now pop from collisions or jumps/falls. The entire wheel can also fall off the car making it much more difficult to drive

When fully repairing the vehicle it will also repair the wheels.


minipunch:
Did u make sure the LS customs repairs the vehicle too ? @Savage912-Official

minipunch:
Nice!! looks good!!

minipunch (approved):

Issue #1217, opened 2022-09-06
try drift smoke script


minipunch:
https://kuzquality.com/package/4906863

Savage912-Official:
Completed in #1220

minipunch:
add

PR #1220, opened 2022-09-06
feat(DriftSmoke): Better Drift Smoke for Vehicles

Savage912-Official:
This script will take your drifting and overall driving experience to the next level. By adding more realistic smoke drifting will feel more alive than ever.

Amazing realistic smoke
Works with colored tire smoke
Smoke only works on the right surfaces. There won't be smoke while offroading.

Issue: #1217 Completed


minipunch (approved):
🔥

PR #1222, opened 2022-09-06
Pressure washer

Rezitoo:
Adds Pressure washer located at many of the gas stations around San Andreas that allow people to swap between water and soap (Middle Mouse Click) while cleaning the persons vehicle.


Savage912-Official:
Nice!

minipunch:
how does it work?

hopefully if i just go to the location it will explain it to me ?

Rezitoo:
> how does it work?
> 
> 
> 
> hopefully if i just go to the location it will explain it to me ?

You walk up and are prompted to press E then you can switch between water and soap with middle mouse click then walk back to the machine and press E again to put the wand away. It will also let you know when your car is clean

Rezitoo:
Looking to test there is one at the car wash by the VU and one at mrpd, as well as almost all of the gas stations throughout the entire state

minipunch:
Ok I found it

1) people won't know to press middle mouse most likely as it doesn't say anywhere i don't think
2) i pressed it and the soap didn't really seem to do anything, and my car was cleaned without me using it
3) people are probably not gonna recognize the pressure washer and just do the normal car wash lol, maybe we should remove the automated car washes?

I guess since soap didn't seem to be needed or even do anything special (other than not make noise when sprayed); 1 and 2 aren't so important

but we might wanna do 3

then again maybe we want to wait on 3 until we make it easier to notice the pressure washer lol

other wise folks will be confused

so for now I'll just merge this as is and we can have both, and u or whoever can tweak as we feel needed on this

minipunch (approved):
Nice, script works good (just a lil hard to find for those who don't know about it probably)

PR #1230, opened 2022-09-08
Repair Kits fix Wheels

Savage912-Official:
Repair kits will not fix the vehicles tires.


minipunch:
Sounds like they need a TOW TRUCK lul
To a mechanic shop
Idk how a roadside repair of ur wheel completely breaking off would go 
Wasn't really intended
But maybe if enough think it's really needed

PR #1236, opened 2022-09-10
Fixed issue with a 911 call

DatBoiWeep:
Suspicious injury calls were not showing up. `.dropEvidence` kept returning `nil` for some reason and couldn't figure it out. So as a temp fix, simply just made a function for suspicious calls to display true if sussy.


minipunch (approved):
👍

minipunch:
Do we know if this has been working or not just to be sure on live ? @DatBoiWeep

DatBoiWeep:
Yes sir and she works like a charm

DatBoiWeep:
image

minipunch:
Niceeee

PR #1240, opened 2022-09-11
fix(AutoRepair): Auto Repair Fix

Savage912-Official:
The current issue regarding the auto repair is if you dont have the money on you when repairing your car the money still gets put into the business account.


minipunch (approved):
nice

PR #1245, opened 2022-09-13
update(Hotwiring): Added Minigame To Hotwiring

Savage912-Official:
Requested by @ChangyBoi

### DIFF
--- a/resources/usa_vehtheft/cl_hotwire.lua
+++ b/resources/usa_vehtheft/cl_hotwire.lua
@@ -182,7 +182,7 @@ AddEventHandler('veh:hotwireVehicle', function(callback, circles)
       end
 
       if not isHotwiring then
-        exports.globals:notify("The number shown is the number than needs to be pressed for the hotwire!")
+        exports.globals:notify("Press the key shown on screen!")
         isHotwiring = true
         if math.random() < HOTWIRE_BREAK_CHANCE then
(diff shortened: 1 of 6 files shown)


minipunch (commented):
Is it just me or if u 1) lockpick a local car and 2) start hotwiring but don't do anything

the minigame goes away but you're stuck in the animation AND can't even move or get out of the car lol?

minipunch (changes requested):
Also can we add some sort of notification that tells you what buttons to press / what to do when the mini game begins

because I have no idea

resources/usa_vehtheft/cl_hotwire.lua
```diff
@@ -182,6 +182,7 @@ AddEventHandler('veh:hotwireVehicle', function(callback, circles)
       end
 
       if not isHotwiring then
+        exports.globals:notify("The number shown is the number than needs to be pressed for the hotwire!")
```
minipunch, line 185:
```suggestion
        exports.globals:notify("Press the key shown on screen!")
```

minipunch:
still same problem

minipunch (approved):
Nice!!

minipunch:
May be a bit easy though

Perhaps we can tweak it to make a bit harder (preferably some random difficulty level each time)

Will push for now since it is still better than before regardless 👍

PR #1252, opened 2022-09-14
update(Hotwire): Decreased hotwire kit break chance

Savage912-Official:
Changed the Hotwire Kit break chance from 0.70 to 0.40


minipunch:
@ChangyBoi just made one already lol

but ty for the PR

PR #1264, opened 2022-09-15
Fixed boosting ped at store

DatBoiWeep:
Now thou shall spawn and delete with a larger radius too


resources/rahe-boosting/api/client.lua
```diff
@@ -165,27 +165,34 @@ end)
...
+        if Vdist(ped.x, ped.y, ped.z, playerCoords.x, playerCoords.y, playerCoords.z) < 100 then
+            if not createdJobPeds then
+                local hash = -984709238
                 RequestModel(hash)
-                Wait(0)
+                while not HasModelLoaded(hash) do
+                    RequestModel(hash)
+                    Wait(0)
+                end
+                local ped = CreatePed(4, hash, ped.x, ped.y, ped.z, ped.heading, false, true)
```
minipunch, line 180: Can we also call `FreezeEntityPosition` on the ped so they don't run away or smth plz

minipunch (changes requested):

resources/rahe-boosting/api/client.lua
```diff
@@ -165,27 +165,34 @@ end)
 
 -- Create Store Ped
 Citizen.CreateThread(function()
-    local spawn = 0
-	while spawn == 0 do
+    local createdJobPeds
```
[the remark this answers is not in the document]
DatBoiWeep: Yes dad

resources/rahe-boosting/api/client.lua
DatBoiWeep: yes

DatBoiWeep:
Fixed

minipunch (approved):

PR #1265, opened 2022-09-16
[POLISHING & TESTING] High Phone

maxgestic:
New Phone system
- Online Banking
- App Store
- Dark Chat
- Group Chats
- Anonymous Phone Calls
- Camera
- Better Messaging
- Calling 911
- Better experience using images
- Old Number and Contacts "should" be moved accross


minipunch (commented):
Can u notify the user they need to go to general store to get phone if they don't have one when using the phone hotkey

minipunch:
Also seems like you can't open the phone when downed?

maxgestic:
> Can u notify the user they need to go to general store to get phone if they don't have one when using the phone hotkey
> 
> u can use the message gcphone uses now

Err I can do that!

maxgestic:
> Also seems like you can't open the phone when downed?

Yea this is true, but then again if your passed out realistically you can't use your phone

minipunch:
> > Also seems like you can't open the phone when downed?
> 
> Yea this is true, but then again if your passed out realistically you can't use your phone 

Ye but being downed doesn't always mean ur "passed out"

U could have gotten into a fist fight with some random and if they leave u after downing you how are you supposed to get help lol

And realistically u might be able to call someone for help in that scenario

minipunch:
1) It says another currency when receiving money in the notification I got when the phone was open lol

2) I replied to one tweet by @Rezitoo but it left my reply on ALL the tweets lol (there was 3 tweets and it replied to all of em instead of just the 1 I wanted to reply to)

minipunch:
3) it also seems like you have to make a contact to call someone? can't dial it? is this what u said the author was working on a fix for?

Rezitoo:
https://i.imgur.com/7Egry1l.png this is the message

minipunch:
the tweet things is cause we are using MySQL v8 instead of v5 and since I removed the default values for tweet 'likes', saving fails because the DB throws a no default value error when executing the query

minipunch:
Well just allowing null values seems to work okay

but now I get 

```[high_phone] An error happens for query "INSERT INTO phone_tweets ... ER_TRUNCATED_WRONG_VALUE_FOR_FIELD: Incorrect string value: '\xF0\x9F\xA4\xA8' for column 'content' at row 1```

When trying to save any tweets with emojis lol...

minipunch:
Also r u able to reproduce the tweets on saving on server load @maxgestic  ? Seems to be the case now even though it seems to be saved to the database

minipunch (approved):
🔥🔥🔥

PR #1270, opened 2022-09-17
update(Insurance): Pay with bank rather than cash

Savage912-Official:
When at the insurance place you no longer need to have cash on you as it will now take it from the bank.


minipunch (approved):
👍

PR #1285, opened 2022-09-20
Fixed NULL name for custom vehicles

DatBoiWeep:
**Issue**
Vehicles display NULL on police calls, boosting, racing, etc.

**Mainly for Northpeak**
Anytime you add a new vehicle, simply open the Vehicle's vehicle.meta and look for <gameName>. (Also can be spawn model)
Then you can name it to anything you like on the second part of the AddTextEntry

**Example**
image

**Was there an easier way?**
Probably, but this was the only way I found LOL


minipunch:
Did u wanna update `server.cfg` to start this new resource

DatBoiWeep:
> Did u wanna update `server.cfg` to start this new resource

LOL Good point. Fixed

minipunch (approved):

PR #1295, opened 2022-09-21
Two new garage locations 

Rezitoo:
Added two locations one near PDM next to bennys mlo and in the parking lot of benefactor


minipunch:
Did u test these in game on ur test server to make sure the blip isn't floating or missing or anything like that

Rezitoo:
> Did u test these in game on ur test server to make sure the blip isn't floating or missing or anything like that

Yessir all good

minipunch (approved):

PR #1303, opened 2022-09-22
Prevent Electric Cars from using NOS

Urbaniized:


minipunch (approved):

resources/usa_nitro/client/utils.lua
```diff
@@ -12,6 +12,19 @@ local ELECTRIC_VEHICLES = {
   [GetHashKey('SURGE')] = true,
   [GetHashKey('VOLTIC')] = true,
   [GetHashKey('TEZERACT')] = true
+  [GetHashKey('I8')] = true
```
minipunch, line 15: No commas???

This means u didn't test 😡😡😡

minipunch (changes requested):

resources/usa_nitro/client/utils.lua
Urbaniized: uhhh... 

I actually did test and it gave no errors and the script worked fine xD

Let me add them

minipunch (approved):

PR #1304, opened 2022-09-22
Added untracked yarn file cos it keeps being there in git status

maxgestic:
Its just a file thats annoying me


minipunch:
I think we should probably add it to the `.gitignore` file instead for now

minipunch (changes requested):
see comment

PR #1305, opened 2022-09-22
Changed Euro to Freedom Dollar

maxgestic:
Changed notifications so Euro is now USD


minipunch:
Nice

Don't forget about changing the displayed name from steam -> char name too

minipunch (approved):

PR #1311, opened 2022-09-23
Added Disconnect to Rockstar Editor to put player in solo session to prevent crashes

maxgestic:


minipunch (approved):
Where u get this idea from ?

PR #1328, opened 2022-09-27
feat(Megaphone): Megaphone / Vehicle Loudspeaker

Savage912-Official:
Usage on Foot: Left Mouse Button
Usage in Cop Car: (~)
Change Megaphone Volume: Mouse Scoll Wheel

Megaphones can be bought from the Electronics section of the Hardware Store.

*If in a police vehicle you don't have to have the megaphone item for the vehicle loudspeaker*


minipunch (approved):
👍

PR #1329, opened 2022-09-28
update(Cams) Move Prison Security Cams to Groups

Crownapple:
I moved the prison security camera into groups due to them being more than 20 cams and its hard to find the one you need.


minipunch:
Did you test this to make sure it still worked ? The diff looks a bit strange

Crownapple:
> Did you test this to make sure it still worked ? The diff looks a bit strange

Yes all cameras where tested

minipunch (approved):

PR #1339, opened 2022-09-29
Fixed Darkchat to actually work proper

maxgestic:
Fixed a before issue of my implementation its better now and has working notifications and all that working as it was meant to


minipunch (commented):
What was wrong with it ? I never noticed any issues with DC

maxgestic:
> What was wrong with it ? I never noticed any issues with DC

I did a lazy fix for its sync not to work which basically every time any person sent a darkchat message it would update everyones darkchat

minipunch:
> > What was wrong with it ? I never noticed any issues with DC
> 
> I did a lazy fix for its sync not to work which basically every time any person sent a darkchat message it would update everyones darkchat

Hm ok.. I don't really get it still but I'll merge it lol

minipunch (approved):

Issue #1331, opened 2022-09-28
mute downed players


minipunch:
because lots of players can't control their emotions after being downed in a video game so they feel the need to go OOC and just verbally harass their downer instead of RPing their injuries

so for now, like the good old days of ARMA II, we'll go ahead and make sure downed folk can't use their mics anymore

EMS will still get their RP in (and perhaps more of it without needing to deal with annoying players who can't RP well) and will still be notified without issue due to the 911 calling system not requiring voice when downed

should also help with players being banned (most times I have seen is when a downed player is mad and lets a 'no no word' slip so some admin bans them)

minipunch:
@Savage912-Official said he would try this

minipunch:
Savage got stuck so I'll take a gander

minipunch:
Sounds like @maxgestic is on the case 🔍

minipunch:
added

PR #1342, opened 2022-09-29
Mute downed

maxgestic:
Downed Players will now not be able to speak as of request of Mini https://github.com/minipunch/usarealismrp-fxserver/issues/1331


resources/mumble-voip/client.lua
```diff
@@ -308,21 +309,23 @@ function MuteVehiclePassengers(playerData)
...
+		local newMode = modeIndex
 
-	if newMode > #mumbleConfig.voiceModes then
-		voiceMode = 1
-	else
-		voiceMode = newMode
-	end
-	
-	NetworkSetTalkerProximity(mumbleConfig.voiceModes[voiceMode][1] + 0.0)
+		if newMode > (#mumbleConfig.voiceModes - 1) then
```
minipunch, line 318: Why the `- 1` here?

resources/mumble-voip/client.lua
```diff
@@ -852,20 +871,24 @@ Citizen.CreateThread(function()
...
-					else
-						voiceMode = newMode
-					end
+					if not disabledVoice then 
+						local voiceMode = playerData.mode
+					
+						local newMode = voiceMode + 1
 					
-					NetworkSetTalkerProximity(mumbleConfig.voiceModes[voiceMode][1] + 0.0)
+						if newMode > (#mumbleConfig.voiceModes - 1) then
```
minipunch, line 879: same question here

resources/usa_death/client.lua
```diff
@@ -113,6 +114,8 @@ Citizen.CreateThread(function()
 		local ped = GetPlayerPed(-1)
 
 		if (IsEntityDead(ped)) then
+			wasDead = true
+			TriggerEvent("mumble:setDead", true)
```
minipunch, line 118:
```suggestion
                        if not wasDead then
			  TriggerEvent("mumble:setDead", true)
                        end
```

minipunch: and just move the `wasDead = true` part to after we call the event

minipunch: this way we only call it once as needed and not every frame right

resources/mumble-voip/client.lua
maxgestic: Because I added Dead as a new "voice mode" i dont want people to be able to switch to Dead via F2, so this limits it to just shout, whisper and normal

resources/usa_death/client.lua
maxgestic: True

minipunch (approved):

minipunch:
Nice

PR #1353, opened 2022-10-02
[EXAMPLE] Removed Tony's card from trading cards

DatBoiWeep:
Removed Tony's trading card from being found in house robberies


DatBoiWeep:
Showing @Khajit71 how to use GitHub

Issue #1351, opened 2022-10-02
new firetruck triggering weapon damage modifier flag


minipunch:
for some reason lol

Khajit71:
this flags because of the weapon damage modifier anticheese. It flags for any vehicle that can shoot something, such as the tanks, buzzard, plane with guns,etc. So maybe it can be made to ignore some specific vehicles?

PR #1354, opened 2022-10-02
fix(Anticheat): Firetruck Flag Fix

Savage912-Official:
Issue: #1351


minipunch:
Gonna mark as a draft since not tested

Savage912-Official:
Tested and seems to be working as intended.

resources/usa_rp2/client/cl_weaponDamageModifier.lua
```diff
@@ -78,21 +78,26 @@ end)
 -- 'modified weapon damage' anti cheat
 Citizen.CreateThread(function()
   local lastCheck = 0
+  local me = PlayerPedId()
+  local veh = GetVehiclePedIsIn(me, false)
```
minipunch, line 82: Calling natives (like `GetVehiclePedIsIn`) is performance intensive, and doing it every other frame (i.e., `Wait(1)` in a while true loop) is not good for optimization

Let's only call it when we need it 

So move it down to like under the check on like 85 so it gets called less and only when required

minipunch (changes requested):

resources/usa_rp2/client/cl_weaponDamageModifier.lua
Savage912-Official: Working on this now

Savage912-Official: Submitted change

minipunch (commented):
Did u actually test to confirm this worked?

Cuz I am wondering if line 100 is actually the line being called when throwing the flag or if this will actually fix it since that alert is thrown in more than just the lines u changed

resources/usa_rp2/client/cl_weaponDamageModifier.lua
minipunch: Looks better nice

minipunch (approved):

PR #1355, opened 2022-10-02
Updated welfare check

Khajit71:
Updated paycheck to $100, due to the high inflation that we have been experiencing lately, as well since all paychecks went up


minipunch:
Don't think we need to right now

PR #1358, opened 2022-10-03
Unconscious Rework

maxgestic:
- If downed you are unable to speak local and on radio (your character will be "unconscious" depending on how many serious injuries you have
- Emergency Services are able to use /epi {id} to inject a downed player with epinephrine to allow them to speak for 5 minutes while in the downed state (if revived and the player dies again the epi resets and the player will have to be re-injected to be able to speak again)


resources/rp-radio/client.lua
```diff
@@ -526,7 +526,7 @@ Citizen.CreateThread(function()
         local broadcastType = 3 + (HAS_EARPIECE and 1 or 0) + (radioConfig.AllowRadioWhenClosed and 1 or 0) + ((Radio.Open and radioConfig.AllowRadioWhenClosed) and -1 or 0)
         local broadcastDictionary = Radio.Dictionary[broadcastType]
         local broadcastAnimation = Radio.Animation[broadcastType]
-        local isBroadcasting = IsControlPressed(0, radioConfig.Controls.Broadcast.Key) and not IsPlayerFreeAiming(PlayerId()) and GetLastInputMethod(0) and not IsEntityDead(playerPed)
+        local isBroadcasting = IsControlPressed(0, radioConfig.Controls.Broadcast.Key) and not IsPlayerFreeAiming(PlayerId()) and GetLastInputMethod(0) and exports.usa_injury:isConscious(GetPlayerPed(-1))
```
minipunch, line 529: Careful with calling exports in a while loop like this is being called now: exports are 'expensive' to call (I think due to it having to search for the function to use) so this will most def raise the cpu msec usage (seen via `resmon`)

U should double check resmon to confirm that is indeed the case

And a fix would be to 'cache' the desired export function here locally and use it instead

So something like:

```
local injuryScript = exports.usa_injury

-- then call it like this:
if injuryScript:isConscious(GetPlayerPed(-1)) then
```

resources/usa_death/client.lua
```diff
@@ -93,6 +93,13 @@ function respawnPed(ped,coords)
 	TriggerEvent("chatMessage", "", { 0, 0, 0 }, "^1^*[RESPAWN] ^r^7You wake up at the local hospital, and can't seem to remember events leading up to now...")
 end
 
+RegisterNetEvent("usa_death:epi")
+AddEventHandler("usa_death:epi", function(bool)
+	if IsEntityDead(GetPlayerPed(-1)) then
```
minipunch, line 98:
```suggestion
	if IsEntityDead(PlayerPedId()) then
```
I think that is a cleaner way to get the cur ped handle

resources/usa_death/client.lua
```diff
@@ -114,10 +121,12 @@ Citizen.CreateThread(function()
 		local ped = GetPlayerPed(-1)
 
 		if (IsEntityDead(ped)) then
-                        if not wasDead then
-			  wasDead = true
-			  TriggerEvent("mumble:setDead", true)
-                        end
+            if not wasDead then
+				wasDead = true
+				if not exports.usa_injury:isConscious(GetPlayerPed(-1)) then
```
minipunch, line 126: if u used the cache method above make sure u change how u call this also

so it would become: ``if not injuryScript:isConscious(PlayerPedId()) then``

resources/usa_injury/__resource.lua
```diff
@@ -4,5 +4,6 @@ client_script "cl_injury.lua"
 server_script "sv_injury.lua"
 
 exports {
-    "getPlayerInjuries"
+    "getPlayerInjuries",
+    "isConscious"
```
minipunch, line 8: fyi in the future an alternative/easier way to define exports if u didn't already know is in some .lua file other than fxmanifest.lua or __resource.lua (can be the same file the func is defined in):

```
exports('exportNameHere', function()
 -- function code for inline function, u can also just pass the name of an already defined elsewhere function instead
end)
```

resources/usa_injury/cl_injury.lua
```diff
@@ -178,6 +178,71 @@ AddEventHandler('injuries:showMyInjuries', function()
...
 end)
 
+local isEpi = false
+local epiTimer = nil
+
+Citizen.CreateThread(function()
+    while true do
+        Citizen.Wait(1000)
+        while isEpi == true do
+            Citizen.Wait(1000)
```
minipunch, line 188: Should use `GetGameTimer()` for measuring time instead of `Wait()` since wait varies based on the client's current FPS, so it might take longer and might take shorter whereas GetGameTimer is always accurate

minipunch: U can see lots of other scripts use it if u need an example

minipunch: Also what's with the double while loop lol, might be able to get rid of that also maybe if u use the above method

minipunch (changes requested):
Looks good but just check out my comments

resources/rp-radio/client.lua
maxgestic: Done

resources/usa_injury/__resource.lua
maxgestic: Fair

resources/usa_injury/cl_injury.lua
maxgestic: Done

minipunch (approved):

PR #1363, opened 2022-10-04
Reduced slots

Khajit71:
Reduced the slots from 150 to 100 | Would look better when the server has like 70/100 rather than 70/150. Better for new players, makes the server look more full, and more appealing.

**COULDN'T TEST BECAUSE MY PC WOULD IMPLODE IF I WERE TO RUN A SERVER WITH 100 SLOTS!!!**


Khajit71:
@minipunch

minipunch:
There is a max player count variable in the `connectqueue` resource that needs to be changed to match the new count also

minipunch (changes requested):
see comment

minipunch (approved):

PR #1371, opened 2022-10-08
Fire Script

maxgestic:
Random Fires for EMS to fight, can also be created by staff via /startfire


minipunch:
Nice

Does it create fires every 30 minutes ?? I quickly looked but it didn't appear to have any sort of 'chance' every 30 mins so I think it's 100% every 30 min a fire will start?

If so that might be a bit TOO many fires

minipunch:
Also I assume you made it send some sort of 911 notification for EMS with a location right?

maxgestic:
> Also I assume you made it send some sort of 911 notification for EMS with a location right?

Yes

maxgestic:
> Nice
> 
> Does it create fires every 30 minutes ?? I quickly looked but it didn't appear to have any sort of 'chance' every 30 mins so I think it's 100% every 30 min a fire will start?
> 
> If so that might be a bit TOO many fires

Fixed here https://github.com/minipunch/usarealismrp-fxserver/pull/1371/commits/643ed9f8f91da8e82f3a22a5ca93daeebbb55433

minipunch:
Still think we might need to end up adjusting that fire count but maybe not, so will merge for now assuming it all works for me

minipunch:
Hm not sure what I am doing wrong but I do /createfire and no fire starts

minipunch:
Oh lmao my bad, it's /startfire not /createfire like u said

minipunch:
blips look like they are stacking which would be nice to get fixed:

minipunch:
Also what about the fire HOSE part of it ?

Does that work?

minipunch:
Nvm, seemed pretty self explanatory, just go to the truck and press E to pick up / drop it for the most part

I think there is some 'refilling' involved but idk

minipunch:
Looks 🔥 for the most part

minipunch (approved):

PR #1375, opened 2022-10-09
Increased item drop chances for dumpsters

Khajit71:
as the title says increased it as it was way too low


minipunch (commented):
Not sure I would agree that this needs raising

Khajit71:
I was dumpster diving today, like 7/10 gave nothing

PR #1385, opened 2022-10-11
Removed more weed spots

Khajit71:
Removed all weed spots other than grapeseed and Best Buds so that people RP more.
Also discussed in Development-discussion and everyone seemed to like it.

**I HAVE TESTED IT**


minipunch:
Again not really keen on this

Khajit71:
Only reason why I have made it, is as we have agreed in dev chat as staff(or staff chat can't remember) to remove them so that more ppl could RP together

PR #1395, opened 2022-10-13
Vehicle Boosting Adjustments

DatBoiWeep:
- Adjusted XP values and Cash rewards across all classes
- Added LicensePlates to the 911 calls
- Added Boosting Vehicles to usa_911calls


DatBoiWeep:
Also mini, pls don't forget to do `npm install` and `npm run build`
Under `rahe-hackingdevice/web`

resources/rahe-boosting/api/client.lua
```diff
@@ -142,7 +142,8 @@ end)
 AddEventHandler('rahe-boosting:client:importantBoostStarted', function(location, vehicleId, vehicleClass)
     local streetName, crossing = GetStreetNameAtCoord(location.x, location.y, location.z)
     streetName = GetStreetNameFromHashKey(streetName)
-    TriggerServerEvent('911:call', location.x, location.y, location.z, "^3Grand Theft Auto^0: OnStar Services reports GPS tampering on vehicle. Vehicle is a [^3"..vehicleClass.."^0] Class. Spotted near ["..streetName.."]")
+    local licenseplate = GetVehicleNumberPlateText(vehicleId)
+    TriggerServerEvent('911:VehicleBoosting', location.x, location.y, location.z, streetName, licenseplate, vehicleClass)
```
minipunch, line 146: Can we make this some chance of calling instead of 100% ?

DatBoiWeep: I guess sir lol. Just keep in mind, this 911 only affects `A` and `S` Class boosts, every other class is just a hotwiring / lockpicking call.

So Ideally, no one would see this 911 till they get to `A` class

minipunch: Yeah idk a call each time might be a lil too OP even though I think but maybe not idk for sure

We can try it for now if u want the way it is

minipunch (approved):

PR #1398, opened 2022-10-13
update(Injury): Allow staff & police inspect cmd

Savage912-Official:
Added the ability for staff to use the inspect command rather than just police.

Requested by: @Khajit71


minipunch:
Why @Khajit71  ?

Khajit71:
Because plenty of times we get /helps from ppl that their injuries got glitched and they need heal, and we need to double check with their injury log if injury's match. That, or if we need to check someone's injuries sometimes we don't have cop staff on @minipunch

Savage912-Official:
@minipunch

resources/usa_injury/sv_injury.lua
```diff
@@ -79,13 +79,19 @@ end, {
...
 	local targetSource = tonumber(args[2])
-	if targetSource and GetPlayerName(targetSource) then
-		TriggerEvent('injuries:getPlayerInjuries', targetSource, _source)
-	else
-		TriggerClientEvent("injuries:inspectNearestPed", _source, _source)
+	local user = exports["essentialmode"]:getPlayerFromId(source)
+	local char = exports["usa-characters"]:GetCharacter(source)
+	local job = char.get("job")
+	local group = user.getGroup()
+	if job == "corrections" or job == "sheriff" or job == "cop" or job == "ems" or group == "mod" or group == "admin" or group == "superadmin" or group == "owner" then
```
minipunch, line 89: Looks like we lost 'doctor' from here ?

minipunch (changes requested):

resources/usa_injury/sv_injury.lua
Savage912-Official: Fixed

minipunch (approved):

Issue #1394, opened 2022-10-13
keybind for manual transmission shifting


minipunch:
for LSHIFT and R

minipunch:
also while looking at this maybe I can take a look at the issue @Urbaniized has been trying to figure out (see DMs)

minipunch:
so @Urbaniized says that basically it only works fresh after a mech installs it

but if u leave ur car or store it basically it goes away at that point

minipunch:
Yeah but if I start the manual transmission script and then pull out the car and get out then it messes the top speed up
And I get some very small number lol
1.40e^-45

minipunch:
going to try reloading what the top speed modifier should be re applying it in the reset vehicle function and see if that works for now

minipunch:
actually just shifting into 1st from neutral seems to remove the top speed? lol

minipunch:
should:

- [ ] only execute `resetvehicle` function body when `manualon` is true to avoid messing up non manny tranny converted vehs to start
- [ ] do what I said above about reloading what it should be set to but only when `manulon` was true

minipunch:
well pushed the first thing, that should fix non manny tranny converted cars at least for now

and we will see if manny tranny converted cars are still missing top speed or not given some time

minipunch:
https://github.com/minipunch/usarealismrp-fxserver/commit/7288dd82aa0336bc070c5378fbb182a199170e82

PR #1400, opened 2022-10-14
EMERGENCY Band-Aid fix for now

Rezitoo:
increased top speed percentage to bring cars back to old top speed. Tested 11 Cars all going previous speed just until we can narrow down the root of the issue.


minipunch:
Yeah don't think this will do it

See my comments on #1394

PR #1423, opened 2022-10-22
update(Gunshop): Added Revolver Back

### LINKED ISSUE #1409: re add revolver
it was disabled since it was a 1 shotter so people felt it was "too OP"

now that we can change damage modifiers for weapons we should re add it and just adjust it so it's like 2-3 hits instead so we can still have the dope revolver

Savage912-Official:
Allow the revolver to be purchased again, and it now takes 3 shots to kill someone rather than 1 shot.

closes #1409


minipunch (approved):

minipunch:
Actually now that I think of this

Instead of just buying it from the gun store

We should make this exclusively craftable as an incentive for blacksmiths

resources/usa-crafting/config.lua
```diff
@@ -192,6 +192,26 @@ Config.recipes = {
             requiredCraftingLevel = 1,
             type = "weapons"
         },
+        { 
+            name = "Revolver",
+            image = "https://wiki.gtanet.work/images/a/a2/weapon_-1045183535.png",
```
minipunch, line 197: U sure this image works?

I can time out error when connecting to that link

can u use imgur instead

minipunch (approved):

minipunch (changes requested):
imgur link comment

resources/usa-crafting/config.lua
Savage912-Official: I'll change it but yeah it worked in game

minipunch (approved):

PR #1424, opened 2022-10-22
update(tebex): added plasma pistol

### LINKED ISSUE #1421: add plasma gun package to store
so players can purchase an exclusive weapon that isn't super OP but still would be dope to have

all while supporting the server

Crownapple:
Added Plasma Pistol to tebex, So able to be added to a tebex package.

closes #1421


minipunch:
thanks for trying this

but I want it as a separate item completely, not just a part of the care package

minipunch (changes requested):
see comment

minipunch:
already did what I requested above, will close this now

PR #1427, opened 2022-10-23
Add Metro and Train System

maxgestic:
- Added Metro and Train Driver Jobs
- Added Drivable Metros and Trains
- Added Ticket System with checks if no valid ticket is present
- Added No Ticket 911 Calls


maxgestic:
Wont work yet cos mini needs to "Buy" the asset since its encrypted

minipunch:
So any little guide on how to start?

Or do I just go to some train station and it will tell me ?

maxgestic:
There are two icons on the map for clocking in, one at the airport subway station and one near the port

minipunch:
Can u try optimizing the client side idle time a bit

As it stands it is in the top 3 of resource-intensive scripts even when not being used

Anywhere ur looping some natives in a `while` and/or `for` loop, often times it can be things like distance checking for 3d text and input handling and what not if there are a lot of items in the collection to loop over

maxgestic:
> Can u try optimizing the client side idle time a bit
> 
> As it stands it is in the top 3 of resource-intensive scripts even when not being used
> 
> Look for anywhere ur calling some natives in a `while` and/or `for` loop, often times it can be things like distance checking for 3d text and input handling and what not if there are a lot of items in the collection to loop over especially
> 
> 

image

minipunch:
Does that pink blip get put on the map for everyone when a train is going around?
Cuz that would be dope if not already doing it lol
But yeah this looks fire
And if you can add NPCs to pick up so players can hop on and do this like trucking to earn money that would be epic
And my usa_trains idles around 0.07 after ur updates, not the 0.03 u have
But that does depend on ur PC/CPU setup so that could just be it
Tho my PC/CPU setup ain't all that bad lmao
I can still prob merge tho for now

minipunch (approved):

PR #1434, opened 2022-10-24
Added Plasma Kart

### LINKED ISSUE #1432: try this plasma kart "mario kart" script
https://forum.cfx.re/t/plasmakart-script-standalone/4790879

looks fire

Crownapple:
Plasma kart is where you can race with your friends in go karts. There is 2 mode regular and battle. Regular is where you guys just race, Battle is where you can race and sabotage your friends by laying spikes and there is boost as well.

CLOSES #1432


minipunch:
I don't think we need the `PataModElevatorV2Lock` one 

It worked fine without that one for me so let's remove that for now since I don't think it's needed?

Savage912-Official:
> I don't think we need the `PataModElevatorV2Lock` one
> 
> It worked fine without that one for me so let's remove that for now since I don't think it's needed?

Yeah, when we tested last night we didn't use that not sure why it got pushed.

Crownapple:
> I don't think we need the `PataModElevatorV2Lock` one
> 
> It worked fine without that one for me so let's remove that for now since I don't think it's needed?

Fixed

Crownapple:
closes #1432

minipunch:
> closes #1432

fyi this needs to go in the PR description

minipunch (changes requested):
@Crownapple needs to work with @ChangyBoi and or @DatBoiWeep to get the clothing component numbers right here for this script

PR #1443, opened 2022-10-25
BOATS!

Rezitoo:
Now you can buy a boat trailer and a addon boat, attach the boat to the trailer then detach it at another spot! easy to add new boats/trailers in the future as well as the default boat trailer


minipunch:
any time u add a new script it would be good if u can show a photo of the `resmon` output for it when idling to make sure it doesn't need to be optimized plz

Rezitoo:
Got you, I'll do that tomorrow

Rezitoo:
@minipunch [https://i.imgur.com/MPcLGgL.png](url) Like this?

minipunch:
Ye looks good

minipunch (approved):

PR #1449, opened 2022-10-27
Add to server.cfg

Rezitoo:
forgot to add script to cfg


Rezitoo:
I don't understand that conflict tbh

minipunch:
that happens when 2 people change the same spot in the same file

so u have to 'resolve' it by telling git what to use from which person

minipunch (approved):

PR #1450, opened 2022-10-27
Prison Evolved Weapons (PEW)

DatBoiWeep:
- Added Anti Air supported by the US Air Force

Process
- Unathorized access entering the prison (or other restricted air access locations) will be prompted with a warning to flee
- After a few seconds, Prison will sound alarm and notify police of the aircraft
- Chance will occur for number of police on duty (100% chance of attack when 0 cops on duty)
- Cooldown will commence to prevent the heli from being repeatedly shot even when already down.
- Emergency Personnel are permitted in the Restricted Area.

**Examples of 911 calls**
image


minipunch:
Well first try nothing happening when I fly solo into the prison alone on the test server

Then after restarting the script it locked the prison down and I saw the 911 messages and warning to leave, but then an F8 error occured, prob cause of restarting it and needing to relog or smth?

minipunch:
Ok well second time trying after restarting the script and relogging I still got the same error:



Also I noticed I only got warnings after I set my job to 'sheriff'

resources/nagodo_antiaircraft/scripts/client.lua
```diff
@@ -0,0 +1,164 @@
...
+                            TriggerEvent("chatMessage", "US Air Force", { 240, 0, 0 }, "RESTRICTED AREA - LEAVE NOW OR BE TAKEN DOWN")
+                            Wait(WaitAlarmTime)
+                            TriggerServerEvent('911:USAF', pCoords.x, pCoords.y, pCoords.z)
+                            if not AlarmTriggered then
+                                Debug("Is Pilot")
+                                inZone = _inZone
+                                TriggerServerEvent("jail:startalarmSV", -1)
+                                Wait(WaitAttackTime)
+                                if not pause then
+                                    if CopsOnDuty >= 1 then
```
minipunch, line 53: this line is throwing the error:

minipunch (changes requested):

resources/nagodo_antiaircraft/scripts/client.lua
DatBoiWeep: This should be fixed now LOL

minipunch:
So it seemed to work better this time after spawning a chopper and flying in as a civ

It notified me and shot me down before I took off to leave

only issue is it threw this error in server console after shooting me: https://i.imgur.com/MHnhSiR.png

minipunch:
Also it seems like all u need is someone in another chopper behind u and it doesn't care about > 1 chopper ?

Is there like a cooldown or smth

minipunch (changes requested):
see comment

DatBoiWeep:
> Also it seems like all u need is someone in another chopper behind u and it doesn't care about > 1 chopper ?
> 
> Is there like a cooldown or smth

There is a cooldown, but this error made the script stop because I never put coords in like a pepega. Should be pushed here in a bit for re-test

minipunch:
Lmao so does it always shoot u down even if it was an accident ??

Can't wait for people to accidentally get shot down lmao

minipunch:
Also we might wanna make it shoot u down a lil bit quicker

resources/nagodo_antiaircraft/scripts/client.lua
```diff
@@ -0,0 +1,170 @@
...
+            if dist <= v.radius then
+                _inZone = true
+                if not IsAllowed(v.allowedJobs) then
+                    if IsPedInAnyHeli(PlayerPedId()) then
+                        Debug("In Heli")
+                        if GetPedInVehicleSeat(GetVehiclePedIsIn(PlayerPedId(), false), -1) == PlayerPedId() then
+                            TriggerEvent("chatMessage", "US Air Force", { 240, 0, 0 }, "RESTRICTED AREA - LEAVE NOW OR BE TAKEN DOWN")
+                            Wait(1000)
+                            TriggerEvent("chatMessage", "US Air Force", { 240, 0, 0 }, "RESTRICTED AREA - LEAVE NOW OR BE TAKEN DOWN")
+                            Wait(WaitAlarmTime)
```
minipunch, line 51: Using `Wait()` is never reliable for time keeping cause it depends on the player's FPS

Should use `GetGameTimer()` instead, see other scripts for examples

resources/nagodo_antiaircraft/scripts/client.lua
```diff
@@ -0,0 +1,170 @@
+--Basic--
+local destroying = false
+local inZone = false
+
+local pause = false
+
+local WaitAlarmTime = 5000 -- 5 seconds
+local AlarmTriggered = false
+local WaitAttackTime = 15000 -- 15 seconds
```
minipunch, line 9:
```suggestion
local WaitAttackTime = 10 * 1000
```

minipunch (changes requested):

resources/nagodo_antiaircraft/scripts/client.lua
```diff
@@ -0,0 +1,170 @@
...
+        SetBlipAsShortRange(blip, true)
+
+        BeginTextCommandSetBlipName("STRING")
+		AddTextComponentSubstringPlayerName(Config.BlipText)
+		EndTextCommandSetBlipName(blip)
+
+    end
+
+    -- Function to get numCops
+    local CopsOnDuty = TriggerServerCallback {
```
minipunch, line 30: Code placed here only runs once when the thread on line 13 is created so this `CopsOnDuty` variable won't be accurate by the time a missile is fired

Should trigger this server callback once just before firing the missile if that is when u need to know how many cops are on instead

minipunch (changes requested):

minipunch:
Pause I think I'm gonna work it around a little and clean it up a lil if possible

I don't really like the use of that `pause` variable in a loop like that lol

minipunch (approved):

PR #1452, opened 2022-10-28
Added new Mechanic clock on spot

Crownapple:
Added a new mechanic clock on spot on greenwich Parkway/ Ls Custom by the airport


minipunch:
U test to make sure it all works well like spawning the truck and the job ped n stuff?

minipunch:
crown didn't test the truck spawn so we'll have to hope it's good since he can't test for a while

minipunch (approved):

PR #1469, opened 2022-10-31
[Staff Commands] Added Halloween Weather & State-wide Blackout

maxgestic:
- Adds ability to set Halloween weather with /weather
 - Ability to toggle blackout for the entire state using /blackout {true/false for blackout} {true/false for cars being blackouted}


resources/realweather/server.lua
```diff
@@ -15,12 +15,13 @@ local WEATHER_TYPES = {
...
     "SNOWLIGHT",
-    "XMAS"
+    "XMAS",
+    "Halloween"
 }
 
 local Data = nil
 
-local cityName = "Los Angeles"
+local cityName = "Los+Angeles"
```
minipunch, line 24: U sure this is gonna work still?

maxgestic: URL encoding can't have spaces it was throwing an error for a while because of this

minipunch: R u *sure* the error was from this though lol

Because p sure it was working fine, the only error I saw was from an undefined 'data' var or smth like dat

maxgestic: Yea cos when the error was happening this fixed it on my test server. I just never pushed it cos it resolves itself but it's probs good to have the plus there cos that's proper URLs

minipunch: Never saw this error before though

Not sure why you would see it but I wouldn't lol

And I wouldn't doubt the space gets handled by the API itself

minipunch: Oh wait u know what, ur right

I just didn't investigate the error I was seeing enough lol

Good fix here

minipunch (approved):
nice

PR #1492, opened 2022-11-08
update(Badge): Update Badge CMD / EMS Ranks

Savage912-Official:
Updated the EMS Ranks to be how they are on the EMS Roster, and added a EMS Badge.

Also updated the badge command for it to use args due to if you are whitelisted for multiple depts you could only use the first badge listed.

EMS Ranks Supplied By @Khajit71

Usage: /showbadge (sasp, bcso, or ems)


minipunch:
> "Also updated the badge command for it to use args due to if you are whitelisted for multiple depts you could only use the first badge listed."

Does it have a fallback so if no command is provided it defaults to the first/only one ?

Savage912-Official:
> > "Also updated the badge command for it to use args due to if you are whitelisted for multiple depts you could only use the first badge listed."
> 
> Does it have a fallback so if no arg is provided it defaults to the first/only one ? Don't want to force users to have to use the arg if they are only whitelisted for one for example

No but I suppose I
Could add it tomorrow

Savage912-Official:
Fallback for no args added.

minipunch (approved):

PR #1497, opened 2022-11-10
feat(Dev): Dev Tools

Savage912-Official:
Updated the coords commands and added some debug commands. These debug commands are only accessible in F8 and by staff members. The coords commands are only able to be used in F8 but can be used by anyone just like before.

**Usage in F8:**

- coords vector3
- coords vector4
- coords heading

- debug
- debug vehicle or veh

**Coords Commands Example:** https://i.imgur.com/NefMvTe.png


minipunch:
Can you provide some screenshots of what the debug commands output looks like

Savage912-Official:
> Can you provide some screenshots of what the debug commands output looks like

Yeah ill get a picture in a bit.

Savage912-Official:
> Can you provide some screenshots of what the debug commands output looks like

/debug command | It also shows object info on models nearby
image


/debug vehicle or /debug veh command
image

Savage912-Official:
> What's with the `[C] Fix Vehicle` and "Z reset number"stuff?
> 
> Doesn't seem like u can use those hotkeys for just a chat command

Just buttons you can use to fix the vehicle / reset the current debug numbers while in the vehicle debug mode.

minipunch (approved):

PR #1498, opened 2022-11-10
feat(Casino): Aviator (Crash Game)

Savage912-Official:
When the round starts, the game gives you a random number that can appear on the screen. You must cash out before you reach the random number given by the game by placing a bet. If you can cash out before you reach the number, you can see the money you have deposited and the money you have earned on the left. You can see how much the number went up in the previous rounds from the top.

In order to access this you must buy a Casino Dongle from a hardware store and have a laptop on you. When having both of these items use the Casino Dongle to access the game.

[Preview 1](https://cdn.discordapp.com/attachments/980523819246698496/980524111375765546/unknown.png)
[Preview 2](https://cdn.discordapp.com/attachments/980523819246698496/980524220142460988/unknown.png)


resources/oe-aviator/server/sv_aviator.lua
```diff
@@ -0,0 +1,42 @@
...
+	
+	if char == nil then
+		--print("Player loading in")
+	else
+		char.removeMoney(math.abs(data))
+	end
+end)
+
+RegisterServerEvent('oe-aviator:addMoneyCallBack')
+AddEventHandler('oe-aviator:addMoneyCallBack', function(data)
```
minipunch, line 22: Gotta find a way to make this not exploitable

Security token is probably the laziest way but that would be the bare minimum for me to at least accept this PR

Other server sided checks are always preferable to the token method

minipunch (changes requested):

resources/oe-aviator/server/sv_aviator.lua
Savage912-Official: Alright, I'll get the security token and look into adding some other things in a bit.

minipunch:
Is this good to review?

Can u resolve the merge conflict if so

Savage912-Official:
> Is this good to review?
> 
> Can u resolve the merge conflict if so

Depends on if you think its secure enough to be pushed or if you think it can still be exploited.

PR #1505, opened 2022-11-11
New Items

Zeeroqt:
you now can get Tape, Electronic Kit, Electronics, Iron, Copper, Glass, and Plastic from searching dumpsters and chopping cars


minipunch:
I don't think I want to add any items that don't have any function, use, or purpose

PR #1513, opened 2022-11-17
Tattoo v2

### LINKED ISSUE #1048: not able to properly preview new tattoos
current script really only allows "previewing" of tattoos if you don't have any you already have and want to keep, otherwise you have to remove ALL tattoos if you decide you don't want one that you applied before checking out

or wait.. u can if u just run outside huh...

so maybe we should just update it anyway since that's a bit of a hack lol

should be easy enough using the fivem-appearance api

maxgestic:
Changed Tattoos to use Fivem Appearance and other small changes

**MAKE SURE TO RUN "yarn build" in the fivem appearance folder after pulling!!!!!!**

Closes #1048


server.cfg
```diff
@@ -401,8 +401,9 @@ ensure gunmetas
 ensure flashbang # Flashbang
 ensure mp_insidetrack
 ensure usa-crafting
-ensure fivem-appearance
+setr fivem-appearance:locale "en"
 setr fivem-appearance:automaticFade 0
```
minipunch, line 405: We might be able to enable this now

This was only disabled because the tattoos were disabled

I think it applies an automatic fade to certain haircuts or smth....

Not exactly sure which cuts to test with without looking more tho lol

minipunch:
Are tats that players have already going to be removed or will it transfer to this new system ?

server.cfg
maxgestic: I'll have a lookie

maxgestic:
> Are tats that players have already going to be removed or will it transfer to this new system ?

They "should" stay...

minipunch:
see DM from me:

> "Can u make it so it removes their clothing, cause I can't see the tats I am trying on lol"

minipunch (changes requested):
see comment

minipunch:
also just noticed I got this error at some point when trying this:

minipunch:
Oh hold on I forgot about yarn build... let me try again..

minipunch:
> see DM from me:
> 
> > "Can u make it so it removes their clothing, cause I can't see the tats I am trying on lol"

This is fixed after running `yarn  build` in fivem-appearance but that F8 error still appears

maxgestic:
> also just noticed I got this error at some point when trying this:
> 
> 

This is fixed by #[64a3c36](https://github.com/minipunch/usarealismrp-fxserver/pull/1513/commits/64a3c36200c18985cb49b4e33db14219fce69071)

resources/usa_tattooshop/sv_tattoos.lua
```diff
@@ -1,170 +1,29 @@
...
-  return total_cost
-end
-
-function comma_value(amount)
-	local formatted = amount
-	while true do
-		formatted, k = string.gsub(formatted, "^(-?%d+)(%d%d%d)", '%1,%2')
-		if (k==0) then
-			break
+local TATTOO_FEE = 1000
```
minipunch, line 4:
```suggestion
local TATTOO_FEE = 2000
```

minipunch (approved):

PR #1521, opened 2022-11-19
Fixed Snowball Damage Mod Issues

### LINKED ISSUE #1517: snowballs messing up damage modifier?
was reported by staff that anyone who throws a snowball has messed up damage modifiers afterward

makes sense cause it does mess with ur modifier, should check it out

maxgestic:
Fixes #1517


resources/usa_rp2/client/cl_weaponDamageModifier.lua
```diff
@@ -78,7 +78,8 @@ Citizen.CreateThread(function()
       if not lastSelectedWeapon or lastSelectedWeapon ~= currentSelected then
         lastSelectedWeapon = currentSelected
         whatItShouldBeModifier = (WEAPON_DAMAGE_MODIFIERS[currentSelected] or DEFAULT_WEAPON_DAMAGE_MODIFIER)
-        SetWeaponDamageModifier(currentSelected, whatItShouldBeModifier)
+        SetPlayerWeaponDamageModifier(PlayerId(), whatItShouldBeModifier)
+        -- SetWeaponDamageModifier(currentSelected, whatItShouldBeModifier)
```
minipunch, line 82: Can u make sure melee damage is still what it used to be (should be more than just 2-3 punches to down someone)

I had to use `SetWeaponDamageModifier` for that reason, since if I recall correctly when making this `SetPlayerWeaponDamageModifier` doesn't seem to work for melee like fists

minipunch: If it's broken with this change as I expect it to be can you just add `WEAPON_SNOWBALL` to the weapon modifier list and see if it works that way instead

minipunch (changes requested):

resources/usa_rp2/client/cl_weaponDamageModifier.lua
minipunch: Also it looks like it conflicts the the thread defined below this one

That one sets `SetPlayerWeaponDamageModifier` to `1.0`

Yet yours is also setting it, but to something != `1.0`, so I would actually guess that this throws _anticheese flags too if u enable it

maxgestic: oh damn yea ok ill have a look probs just use this custom snowball https://github.com/CookayGaming/CustomSnowball

minipunch: Sounds good

minipunch (changes requested):
see comment

minipunch (approved):

PR #1522, opened 2022-11-19
Limit Business Owning

maxgestic:
Limit Business Amount per Char, current is set at 3 can be changed in sv_config.lua of usa-businesses with the condig value of MAX_OWNED_BUSINESSES


resources/usa-businesses/sv_businesses.lua
```diff
@@ -1,3 +1,5 @@
+BUSINESS_AMOUT_CHAR = {}
```
minipunch, line 1: Would like to see these variables scoped, aka add `local` in front of it so it limits scope to the file it is defined in, unless it is critical that it is defined across files

minipunch (approved):
See comment for future reference, will still merge this for now

PR #1523, opened 2022-11-20
[Map] Ability to toggle groups of Map Blips

maxgestic:
This adds the ability to toggle different groups of map blips on or off. If off they will not be shown on the map. This will hopefully fix the issue of blips flickering when there are too many blips. 

Commands Added: 
/hideblips [blipGroup]
/showblips [blipGroup]
/listblips

Takes it from this:
image
To This:
image


resources/usa_map_blips/client.lua
```diff
@@ -0,0 +1,60 @@
+mapBlips = {}
```
minipunch, line 1: Why not do this stuff in `usa_blips` lol, don't see the need for a separate resource for it

resources/usa_map_blips/client.lua
```diff
@@ -0,0 +1,60 @@
...
+	for key,group in pairs(mapBlips) do
+		for index,blip in ipairs(group) do
+			SetBlipDisplay(blip.blip, 0)
+		end
+	end
+end)
+
+RegisterNetEvent("usa_map_blips:showAllBips")
+AddEventHandler("usa_map_blips:showAllBips", function()
+	print("go")
```
minipunch, line 48:
```suggestion
	
```

resources/usa_map_blips/server.lua
```diff
@@ -0,0 +1,81 @@
...
+        TriggerClientEvent("usa_map_blips:showBlipGroup", source, group)
+    end, {
+        help = "Show Group of Blips",
+        params = {
+            { name = "Group", help = "Group to show (" .. groupstring ..")" }
+        }
+    })
+end
+
+TriggerEvent('es:addCommand', 'hideallblips', function(source, args, char, location)
```
minipunch, line 26: Think this would be cleaner and easier for the user if we just made this command run if they use `/hideblips` without any extra arguments, the less commands to remember the better

resources/usa_map_blips/server.lua
```diff
@@ -0,0 +1,81 @@
...
+    TriggerClientEvent("usa_map_blips:hideAllBips", source)
+end, {
+    help = "Hide All Hidable Blips",
+    params = {
+    }
+})
+
+TriggerEvent('es:addCommand', 'showallblips', function(source, args, char, location)
```
minipunch, line 34: Same here prob in the other direction

resources/usa_map_blips/server.lua
```diff
@@ -0,0 +1,81 @@
...
+        end
+    end
+    if not alreadyIn then
+        table.insert(grouparray, groupName)
+        if groupstring == "" then
+            groupstring = groupName
+        else
+            groupstring = (groupstring .. ", " .. groupName)
+        end
+        addCommands()
```
minipunch, line 68: Lol why are we calling this here and then redoing all the chat suggestions every time a new blip is added ?

maxgestic: Cos every time a new "group" gets added it needs to be added into the help section of the command

resources/usa_map_blips/client.lua
maxgestic: cos i didnt realise it was there and then i didnt wanna move it, grr

maxgestic: Merged it by moving code from usa_blips into this

resources/usa_map_blips/server.lua
maxgestic: Done

resources/usa_map_blips/server.lua
maxgestic: Done

resources/usa_map_blips/client.lua
minipunch: > Merged it by moving code from usa_blips into this

Hm next time instead of cramming it all into a single file it would prob be good to make new files for each of them, but I guess it's fine for now

resources/usa_map_blips/server.lua
minipunch: I don't really see the need for this `groupstring` thing, if it just to update the help text then there has to be another way to do it without needing to add all the extra complexity

Like 'groupstring' looks like it just a list of all the registered 'groups' which is compiled when all the scripts start (aka like right when the player joins) and then it never changes, right?

Just remove it from the help text and point them to some new command that displays the available groups (aka 'groupstring' I believe) so you don't have to worry about readding them and what not

maxgestic: Changed in https://github.com/minipunch/usarealismrp-fxserver/pull/1523/commits/55b960168daa1fb290c25bf7d8e88962acb1d6e2

minipunch (approved):

PR #1529, opened 2022-11-23
update(Evidence): Only show with Flashlight

Savage912-Official:
Evidence will now only show if you have a flashlight in hand and aiming it.

Tested and works lol.


resources/usa_evidence/cl_evidence.lua
```diff
@@ -439,8 +448,9 @@ Citizen.CreateThread(function()
     local lastCheck = 0
     while true do
         if GetGameTimer() - lastCheck > POLL_INTERVAL_SECONDS * 1000 then
-            local mycoords = GetEntityCoords(PlayerPedId())
-            TriggerServerEvent("evidence:loadNearbyEvidence", mycoords)
+			playerPed = PlayerPedId()
+            playerCoords = GetEntityCoords(playerPed)
+            TriggerServerEvent("evidence:loadNearbyEvidence", playerCoords)
```
minipunch, line 453: why these changes / additions ?

minipunch: what was wrong with `mycoords`

resources/usa_evidence/cl_evidence.lua
```diff
@@ -416,13 +421,17 @@ end)
 Citizen.CreateThread(function()
 	while true do
 		for i = 1, #droppedEvidence do
-			local item = droppedEvidence[i]
-			local text = '[U] - Discard Evidence'
-			if onDuty then
-				text = '[U] - Pickup '..item.string
+			player = PlayerId()
+			playerPed = PlayerPedId()
+			if IsPlayerFreeAiming(player) and GetSelectedPedWeapon(playerPed) == GetHashKey("WEAPON_FLASHLIGHT") then
```
minipunch, line 426: native functions (i.e. `IsPlayerFreeAiming`) are CPU intensive to call so we should limit calling them as much as possible

therefore you should move this if statement check above the `for` loop since you don't need to check for every piece of evidence if they are free aiming.. only once before looping thru all the pieces of evidence..

minipunch (changes requested):

resources/usa_evidence/cl_evidence.lua
Savage912-Official: No reason tbh and ill change back once github clone is done.

resources/usa_evidence/cl_evidence.lua
Savage912-Official: I'll change once github clone is done.

minipunch (approved):

PR #1532, opened 2022-11-24
Fixed Train Dependency Issues

maxgestic:


resources/[system]/globals/cl_global.lua
```diff
@@ -352,4 +352,12 @@ exports("sleep", function(timeMs)
   while GetGameTimer() - start < timeMs do
     Wait(1)
   end
+end)
+
+local isPassanger = false
```
Comment (minipunch) [inline resources/[system]/globals/cl_global.lua:357]: Not a big fan of keeping this train state in the globals resource, seems very out of place

Instead of doing this can we first try what I suggested in dev chat instead?

resources/dpemotes/Client/Emote.lua
```diff
@@ -297,7 +297,7 @@ end
 -----------------------------------------------------------------------------------------------------
 
 function OnEmotePlay(EmoteName)
-  if not exports.usa_trains:checkIsPassanger() then
+  if not exports.globals:checkIsPassanger() then
```
minipunch, line 300:
```suggestion
  if exports.usa_trains and exports.usa_trains.checkIsPassanger and not exports.usa_trains:checkIsPassanger() then
```

minipunch: like that, same applies to the other if statements

minipunch (changes requested):

resources/dpemotes/Client/Emote.lua
```diff
@@ -297,7 +297,13 @@ end
...
 
 function OnEmotePlay(EmoteName)
-  if not exports.usa_trains:checkIsPassanger() then
+  local isPassanger = false
+  if exports.usa_trains then
+    if exports.usa_trains.checkIsPassanger() then
+      isPassanger = true
+    end
+  end
+  if not isPassanger then
```
minipunch, line 306: Can we move all this into a local function... something like ``isPassenger` works fine probably

resources/usa_ammunition/client/cl_ammunition.lua
```diff
@@ -56,34 +56,42 @@ end)
...
-        -- set / remove extended mag if applicable --
-        if type(data) == "table" then
-            if data.magComponent then
-                GiveWeaponComponentToPed(myped, currentWeaponHash, GetHashKey(data.magComponent))
-            else
-                if WEP_EXTENDED_MAG_COMPONENTS[currentWeaponHash] then
-                    for index, hash in pairs(WEP_EXTENDED_MAG_COMPONENTS[currentWeaponHash]) do
-                        hash = GetHashKey(hash)
-                        RemoveWeaponComponentFromPed(myped, currentWeaponHash, hash)
+        if not isPassanger then
```
minipunch, line 66: Same goes for here and other places that follow this same pattern

resources/usa_holster/client.lua
```diff
@@ -138,103 +138,113 @@ end)
...
-						Wait(5)
+		local isPassanger = false
+		if DoesEntityExist(ped) and not IsEntityDead(ped) and not IsPedInAnyVehicle(ped, true) and not exports["usa_stretcher"]:IsInStretcher() then
+			if not exports["usa_stretcher"]:IsInStretcher() then
+		        if exports.usa_trains then
+		            if exports.usa_trains.checkIsPassanger() then
+		                isPassanger = true
+		            end
+		        end
+		        if not isPassanger then
```
minipunch, line 149: Think this is the last place

minipunch (changes requested):

maxgestic:
I had to do it the way its done here because exports.usa_trains returns a table even if the resource is not loaded.

minipunch (approved):

PR #1538, opened 2022-11-26
Updated Phasing to Last Whole Race

Urbaniized:
Lots of people have desync issues and causes racers to constantly crash into each other, especially when more than 8 people are in a single race.


minipunch:
would be ideal if this could be like optional or something

minipunch (approved):

PR #1543, opened 2022-11-27
feat(Vehicles): Flip Vehicle CMD

Savage912-Official:
Ability to flip vehicles with a command.

It takes 25 seconds to flip a vehicle over and you can only do this while outside of the vehicle and close enough to the vehicle.


minipunch:
Not really sure I like this cuz it takes a job away from mechanics / tow + not extremely realistic

Savage912-Official:
> Not really sure I like this cuz it takes a job away from mechanics / tow + not extremely realistic

Alright, good to close this PR?

PR #1548, opened 2022-11-27
Remove Frozen Alamo Sea for Now

maxgestic:


minipunch (approved):

minipunch:
since the frozen lake is HUGE and conflicts with downstairs MLOs like the sandy SO

PR #1574, opened 2022-12-03
[Weather/Xmas] More Snow Chance in December

maxgestic:
In December in addition to the normal 15 minute checks for weather updates, there is now a 3 hourly timer, every 3 hours there is a 50% of snow, if the random chooses snow there is a 2/3 chance for snow on the ground and a 1/3 chance for no snow on the ground only snowfall in air. The snow has a duration for 90 minutes and after which it will return to real weather. It functions the same as if a staff member did `/weather [SNOW/XMAS] 90`


resources/realweather/server.lua
```diff
@@ -25,7 +25,8 @@ local cityName = "Los+Angeles"
 local apikey = "fac7afd04fe5b0747a2b7da0c8b4e2f2"
 local GetWeather = "http://api.openweathermap.org/data/2.5/weather?q="..cityName.."&lang=fr&units=metric&APPID="..apikey
 
-local MINUTES_PER_CHECK = 15
+local MINUTES_PER_CHECK = 15 -- minutes between weather checks
```
minipunch, line 28: not used ?
```suggestion
```

minipunch (approved):

PR #1579, opened 2022-12-04
Lowered Farm values again

Rezitoo:
With the new mins and maxes payout minimum would be 270k max would be 390k, More balance less op


minipunch:
Not sure I want to adjust the farm payouts atm

PR #1580, opened 2022-12-05
Arms fix v3

ChangyBoi:
Re-opened as per request


minipunch (approved):

minipunch:
this should already be in actually

lmk if not next restart

ChangyBoi:
> this should already be in actually
> 
> lmk if not next restart

I shall let you know

PR #1584, opened 2022-12-05
Adjusted resell price of VIN scratch vehicles

DatBoiWeep:
Cause my bois were making hella profit out at the dealerships. So in turn, the pricing of a vehicle will be divided by 100K.

In reality it would be $0 due to dealerships checking the VIN status of the vehicle and would straight up deny that shiz.


resources/rahe-boosting/api/server.lua
```diff
@@ -159,7 +159,7 @@ AddEventHandler('rahe-boosting:server:vinScratchSuccessful', function(playerId,
         hash = vehInfo.hash,
         plate = licensePlate,
         stored = true,
-        price = vehInfo.price,
+        price = vehInfo.price / 100000, -- This will divide actual price for $100K
```
minipunch, line 162:
```suggestion
        price = vehInfo.price * 0.05, -- This will divide actual price for $100K
```

minipunch: How about this instead, just 5% of whatever the price is to avoid decimal numbers and make it not COMPLETELY worthless

minipunch (changes requested):
1) see my comments
2) can u add a security token at the very least for the vehicle reward event so it can't be exploited quite as easily

resources/rahe-boosting/api/server.lua
DatBoiWeep: lmao, fair enough

resources/rahe-boosting/api/server.lua
```diff
@@ -159,7 +159,7 @@ AddEventHandler('rahe-boosting:server:vinScratchSuccessful', function(playerId,
         hash = vehInfo.hash,
         plate = licensePlate,
         stored = true,
-        price = vehInfo.price,
+        price = vehInfo.price * 0.05, -- This will divide actual price for $100K
```
minipunch, line 162:
```suggestion
        price = vehInfo.price * 0.05,
```

minipunch (approved):

PR #1585, opened 2022-12-05
New whips

Urbaniized:


resources/rahe-boosting/config/server.lua
```diff
@@ -583,6 +585,8 @@ supportedVehicles = {
     { name = "Dodge 3500", model = "bcys", class = "B" },
     { name = "Dodge 3500 HD", model = "bc203500hd", class = "B"},
     { name = "Chevrolet C10", model = "bcc10c", class = "B"},
+    { name = "Karin Ariant", model = "ariant", class = "B"},
+    { name = "Karin Asteropers", model = "asteropers", class = "B"},
```
minipunch, line 589:
```suggestion
    { name = "Karin Asterope RS", model = "asteropers", class = "B"},
```

resources/usa_autorepair/cl_autorepair.lua
```diff
@@ -25,7 +25,6 @@ vehicleRepairStation = {
 	{-35.4, -1088.7, 25.55}, -- ls dealership
 	--{-10.37, -1662.44, 28.74}, -- mosley dealership
 	{-780.62, -235.56, 36.25}, -- rpg performance
-	{-32.622695922852, -1066.1967773438, 27.396472930908}, -- gabz_hub on Power St
```
minipunch, line 28: Why remove this?

minipunch (changes requested):

resources/usa_autorepair/cl_autorepair.lua
Urbaniized: It's been switched to an LS customs now, so no need to have a repair spot there anymore. I switched it out for a wheel fitment area

minipunch (approved):

PR #1587, opened 2022-12-06
feat(Heli/Planes): Submix Audio

Savage912-Official:
Makes the helicopters and planes not so loud when you are in them (Kinda makes it like your character has headphones on)


minipunch:
Nice

PR #1612, opened 2022-12-14
update(Misc): Basketball Blip Updates

Savage912-Official:
Made markers not show around the basketball, and hid the blips that weren't needed.


resources/[basketball]/rcore_basketball/config.lua
```diff
@@ -22,7 +22,7 @@ Config = {
 		Flat = false,
 	},
 
-	HideMarkers = false,
+	HideMarkers = true,
```
Comment (minipunch) [inline resources/[basketball]/rcore_basketball/config.lua:25]: I think those markers are probably pretty useful to have on
```suggestion
	HideMarkers = false,
```

Savage912-Official:
Updated

minipunch (approved):

PR #1613, opened 2022-12-14
feat(Misc): Progress Bar Updates

Savage912-Official:
Changed the current progressbar to a progressbar which is built into ox_lib. It is able to be cancelled and the cancel button can be changed in the fivem keybind settings.

Previews are in group chat.

Default cancel key is: BACKSPACE


minipunch (changes requested):
Seems pretty good

But the nice thing about the other one is it tells you what key it is to cancel on screen while ur doing it

Can u add that to this please

Savage912-Official:
> Seems pretty good
> 
> But the nice thing about the other one is it tells you what key it is to cancel on screen while ur doing it
> 
> Can u add that to this please

Yep I’ll do it in the morning.

resources/ox_lib/resource/interface/client/progress.lua
```diff
@@ -122,6 +122,10 @@ local function startProgress(data)
             end
         end
 
+        if progress?.canCancel then
+            TriggerEvent('usa:showHelp', true, 'Cancel Progress Bar: ~INPUT_FRONTEND_RRIGHT~')
```
minipunch, line 126: It looks like there is an extra R in the `INPUT_FRONTEND_RRIGHT` part ?

minipunch: savage said it was how it was meant to be

minipunch (approved):

PR #1614, opened 2022-12-14
update(PoliceHeli): New & Improved Script

Savage912-Official:
New and Improved Heli Script (More Optimized & More Features)

- Camera (Has Night & Thermal Vision) (**Key:** E | **Themal/Night Vision Key**: Right Mouse Button)
- Heli Seatbelt (**Key:** B)
- Hovering System (For Heli Stabilization) (**Key:** J | **Super Hover Key:** G)
- Help Display (Shows Controls / Info) (**Key:** L)


minipunch:
Where u get it from ?

Savage912-Official:
Yeah I sent to him in dms

minipunch:
When u say it is more 'optimized', how so ?

This is the idle time for the current script on live:

minipunch:
idle time in heli with cam running is:

minipunch:
So it is a bit more optimized I guess at idle, the new one looks like it idles at even less: `0.00` which is nice

minipunch:
So now:

1) why the extra/separate seat belt system, why not use the one we have now since it seems to function in the polmav already too, not sure if we need to add it to others or what?

2) think u can add like a little help text that says 'press L for help' when u get into a compatible vehicle (one with a helicam)

minipunch (changes requested):
see comment

Savage912-Official:
> see comment

Completed

minipunch (approved):
nice

PR #1621, opened 2022-12-16
update(Crosshair): New Crosshair

Savage912-Official:
This update adds a new crosshair and removes the old one so when you have your crosshair on someone it doesn't turn red anymore.

Requested by: @ChangyBoi


resources/usa-crosshair/client.lua
```diff
@@ -33,6 +41,26 @@ Citizen.CreateThread(function()
...
+CreateThread(function()
+	while true do
+		Wait(0)
+		if IsPedArmed(PlayerPedId(), 6) then
+			if (IsPlayerFreeAiming(PlayerId())) then
+				aiming = true
+				Wait(100)
+			else
+				aiming = false
+				SendNUIMessage("crosshairHide")
```
minipunch, line 53: Is this necessary to call when `crosshairHide` is already being called above on line 26?

Savage912-Official: I suppose not let me test it out before I change it though

Savage912-Official: So, at this current moment it is needed because if you stop aiming the crosshair stays on the screen until you put your gun away. It's mainly being called on the line above to perform the checks for Config.Weapons.Reticle.

Savage912-Official: I did remove the one right below for the if is ped armed check, since we only needed it for is player free aiming.

minipunch: Hm ok

minipunch: Also not a fan of calling `SendNUIMessage` on every frame as is done on line 53

We should only need to call that once to activate and once to deactivate instead of in a loop like that

minipunch (changes requested):

minipunch (approved):

PR #1622, opened 2022-12-16
update(ContainerJob): Slight Pay Increase

Savage912-Official:
This pay increase makes the container job pay out a little more since alot of the other jobs pay way more.

Requested by: @Crownapple


minipunch:
not sure it needs it

PR #1636, opened 2022-12-20
Less Than Lethal Changes

maxgestic:
- Added Beanbag Shotgun
- Added Effect to player if hit by less than lethals


resources/_dmgevents/cl_damageEvents.lua
```diff
@@ -8,6 +8,28 @@ AddEventHandler(littleTrickToSkipGettingScrambledByParser, function(event, data)
         isMeleeDamage = data[12]
         vehicleDamageTypeFlag = data[13]
 
+        if IsEntityAPed(victim) and IsPedAPlayer(victim) and PlayerPedId() == victim then
+            if weaponHash == GetHashKey("WEAPON_BEANBAGSHOTGUN") then
```
minipunch, line 12: For the sake of optimizing can you instead of getting the hash keys for the weapon every time a damage event is triggered can you do it at the top of the file for example so it only runs once and then just use the value

P sure we have gone thru this on another PR before

minipunch: Throughout the file

resources/_dmgevents/cl_damageEvents.lua
```diff
@@ -8,6 +8,28 @@ AddEventHandler(littleTrickToSkipGettingScrambledByParser, function(event, data)
...
+                Citizen.CreateThread(function()
+                    SetPedToRagdoll(PlayerPedId(), 2500, 2500, 0, 0, 0, 0)
+                    ShakeGameplayCam("FAMILY5_DRUG_TRIP_SHAKE", 1.0)
+                    AnimpostfxPlay("DefaultFlash", 15000, false)
+                    AnimpostfxPlay("BeastTransition", 15000, false)
+                    AnimpostfxPlay("CrossLine", 15000, false)
+                    Citizen.Wait(15000)
```
minipunch, line 19: Again let's not rely on `Wait` for timing things on the client, use GetGameTimer() instead please

minipunch (changes requested):

resources/_dmgevents/cl_damageEvents.lua
maxgestic: Fixed https://github.com/minipunch/usarealismrp-fxserver/pull/1636/commits/f459c0eaaafc8354bacdaa09513b98cb8e793c4f

resources/_dmgevents/cl_damageEvents.lua
maxgestic: Fixed https://github.com/minipunch/usarealismrp-fxserver/pull/1636/commits/94e50f6c5de8b95ea4a0d7238de9dff624dacc44

minipunch (approved):

Issue #1623, opened 2022-12-16
remove /underglow command 


minipunch:
and its associated code

since that RGB controller script added a 'G' hotkey to do the same thing

and we should try to minimize the amount of underglow related code, esp since when we don't need it anymore

minipunch:
will make fixing the disappearing underglow issue easier

and I wonder if that is it alone, idk who wrote that stuff

minipunch:
did it

and it looked very weird indeed

let's see if that was the source of the disappearing underglow now...

PR #1650, opened 2022-12-23
Possible Fix For Neon Issue

Urbaniized:


minipunch:
I don't really see how this would be a good 'fix' for the reasons we chatted about in voice

minipunch:
Let's see if this fixes it first: https://github.com/minipunch/usarealismrp-fxserver/issues/1623#issue-1499403811

PR #1660, opened 2022-12-25
Fixed sell price at dealership issue

DatBoiWeep:
Fixed VIN scracthed vehicles selling at a different price from the stated return value

### DIFF
--- a/resources/usa_carshop/vehicle-shop_sv.lua
+++ b/resources/usa_carshop/vehicle-shop_sv.lua
@@ -1033,7 +1033,13 @@ RegisterServerEvent("vehShop:sellVehicle")
 AddEventHandler("vehShop:sellVehicle", function(toSellVehicle)
 	local usource = source
 	print("toSellVehicle: " .. toSellVehicle.model)
-	local vehiclePrice = GetVehiclePrice(toSellVehicle)
+	local currentPrice = toSellVehicle.price
+	local vehiclePrice
+	if currentPrice > 0 then -- If a price is set above 0 then use existing price
+		vehiclePrice = currentPrice
+	else
+		vehiclePrice = GetVehiclePrice(toSellVehicle) -- If no price exists, use standard sell price
+	end
 	if not vehiclePrice then
 		if toSellVehicle.price then
 			print("vehicle price nil with toSellVehicle: " .. toSellVehicle.make .. " " .. toSellVehicle.model)
--- a/resources/usa_carshop/vehicle-shop_sv.lua
+++ b/resources/usa_carshop/vehicle-shop_sv.lua
@@ -1033,10 +1033,9 @@ RegisterServerEvent("vehShop:sellVehicle")
 AddEventHandler("vehShop:sellVehicle", function(toSellVehicle)
 	local usource = source
 	print("toSellVehicle: " .. toSellVehicle.model)
-	local currentPrice = toSellVehicle.price
 	local vehiclePrice
-	if currentPrice > 0 then -- If a price is set above 0 then use existing price
-		vehiclePrice = currentPrice
+	if toSellVehicle.price then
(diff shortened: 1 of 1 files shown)


resources/usa_carshop/vehicle-shop_sv.lua
```diff
@@ -1033,7 +1033,13 @@ RegisterServerEvent("vehShop:sellVehicle")
 AddEventHandler("vehShop:sellVehicle", function(toSellVehicle)
 	local usource = source
 	print("toSellVehicle: " .. toSellVehicle.model)
-	local vehiclePrice = GetVehiclePrice(toSellVehicle)
+	local currentPrice = toSellVehicle.price
+	local vehiclePrice
+	if currentPrice > 0 then -- If a price is set above 0 then use existing price
```
minipunch, line 1038: why > 0 here ?

minipunch: Is there ever a case where a vehicle has <= 0 for its `.price` attribute?

DatBoiWeep: I put it there in case a car would ever have a value of nil for whatever reason, but I guess you do be right. Would be pointless

minipunch: > I put it there in case a car would ever have a value of nil for whatever reason, but I guess you do be right. Would be pointless

Can we just do `if toSellVehicle.price then` then?

minipunch: And why do some cars not have a .price attached to them in the first place?

DatBoiWeep: All good questions sir. vin scratched will still allow the pricing for it.

DatBoiWeep: I can definitely work on editing it to something else tho

resources/usa_carshop/vehicle-shop_sv.lua
```diff
@@ -1033,10 +1033,9 @@ RegisterServerEvent("vehShop:sellVehicle")
 AddEventHandler("vehShop:sellVehicle", function(toSellVehicle)
 	local usource = source
 	print("toSellVehicle: " .. toSellVehicle.model)
-	local currentPrice = toSellVehicle.price
 	local vehiclePrice
-	if currentPrice > 0 then -- If a price is set above 0 then use existing price
-		vehiclePrice = currentPrice
+	if toSellVehicle.price then
+		vehiclePrice = toSellVehicle.price
```
minipunch, line 1038: This is scary cause like cyber security rule #1 is don't ever trust values sent from a client when doing critical logic stuff like this

Cause someone can trigger this event with say toSellVehicle.price = 10000000000000000 and potentially exploit it that way

Which is why we are using the server sided function `GetVehiclePrice` in the first place, to 'sanitize' the input from the client and get the actual value of what the price should be based on the name or w/e the server says it should be

So how about attaching some new field like `isVinScratched` when the player gets the vehicle from that script and then here we just check `if toSellVehicle.isVinScratched then` and then do `GetVehiclePrice(toSellVehicle) * .10` (10% of the price)

minipunch (changes requested):

resources/usa_carshop/vehicle-shop_sv.lua
DatBoiWeep: So looking into this and why `toSellVehicle.isVinScratched` wasn't returning a value (True/False), it seems once the vehicle is stored it just goes poof, which happens once the vehicle is VIN scratch. We gotta add the VinScratch thingy into the DB values if we want to utilize it the way you intend

DatBoiWeep:
Temporarily closing

PR #1661, opened 2022-12-25
Fixed false anticheese flags for warehouses

### LINKED ISSUE #1654: warehouse heist triggering false anti cheat flags
happens when entering/exiting the warehouse during a heist

DatBoiWeep:
Fixes #1654

Added warp locations into `_anticheese` to prevent spam for staff


minipunch (commented):
did u actually test this on ur test server and confirm it stopped the false alerts?

DatBoiWeep:
> did u actually test this on ur test server and confirm it stopped the false alerts?

yes, I enabled ` _anticheese` and `usa_utils` to make sure it worked

minipunch:
swagged out

minipunch:
I will say a better fix would be to automatically pull all the locations for warp points and load it that way

Otherwise any time we add or remove one we also need to update this separate list :(

minipunch (changes requested):
see comments above, he said he will make an export to get all locations

DatBoiWeep:
**Update:** Tested, turns out we can't use a server export on a client file. So I had to use the client av-warehouse to actually make the export work.

resources/_anticheese/cl_anticheese.lua
```diff
@@ -782,6 +782,12 @@ function isAtAWarpPoint(x, y, z)
 			return true
 		end
 	end
+	local warehousecoords = exports["av-warehouse"]:getWarehouseLocations()
```
minipunch, line 785: exports are CPU intensive to call

so you should move this somewhere where it only runs one time and not every time this function is run

resources/av-warehouse/client/main.lua
```diff
@@ -391,4 +391,12 @@ function alertPolice(coords)
 	local lastStreetHASH = GetStreetNameAtCoord(x, y, z)
 	local lastStreetNAME = GetStreetNameFromHashKey(lastStreetHASH)
 	TriggerServerEvent('911:call', x, y, z, "Warehouse burglary alarm triggered (" .. lastStreetNAME .. ")", "Warehouse Burglary")
-end
\ No newline at end of file
+end
+
+exports("getWarehouseLocations", function()
+	local coords = {
+		[1] = {x = coordx, y = coordy, z = coordz},
```
minipunch, line 398: ????

i don't understand this list of coordinates lol

minipunch (changes requested):

resources/av-warehouse/client/main.lua
minipunch: shouldn't this be the actual list of locations

minipunch: oh right this is using the set coord

maybe it would be better to grab ALL the locations at once instead here

resources/_anticheese/cl_anticheese.lua
DatBoiWeep: ooooooh. Very well, I shall relocate that bad boi

DatBoiWeep:
pls re-examine. It has been tested and works

I also could not for the life of me figure out how to bring the coords from the `server.lua` to the `anticheese` client file as an export.

minipunch:
> I also could not for the life of me figure out how to bring the coords from the `server.lua` to the `anticheese` client file as an export.

just need to use `TriggerServerCallback` to fetch the info inside the get locations export

minipunch:
did it 4 u [245d266](https://github.com/minipunch/usarealismrp-fxserver/pull/1661/commits/245d2660ad53700599b9bd273e47741e5d15429a)

minipunch (approved):

PR #1671, opened 2023-01-02
make 2 Fleeca banks on Hawick Ave robbable 

### LINKED ISSUE #1655: make 2 Fleeca banks on Hawick Ave robbable
as title states we should make them heist-able since rn u cannot for some reason

DatBoiWeep:
Fixes #1655

**Additions**
- Added lock doors to both Hawick Fleeca banks
- Added drillable lock boxes to both Fleeca banks


DatBoiWeep:
image
sheesh, she works

minipunch:
Does drilling the boxes / entering the vault also trigger a 911 alert?

Otherwise people might somehow glitch in and out without anyone knowing?

DatBoiWeep:
> Does drilling the boxes / entering the vault also trigger a 911 alert?
> 
> Otherwise people might somehow glitch in and out without anyone knowing?

Drilling the boxes triggers the 911 call to go through that there is a heist. For the gates, it'll automatically send the call for door being lock picked

minipunch (approved):

PR #1674, opened 2023-01-03
Add Hotwiring Kits to Crafting Table

Urbaniized:


Urbaniized:
-Added Hotwiring Kits to Crafting table - Craft Time 3 minutes
 *Requires 3x Plastic 6x Copper Wire
-Added Crude Oil to Mining (rare)
-Players can refine 1x Crude Oil to get 1x Plastic & 2x Copper to get 1x Copper Wire
-Plastic and Copper Wire can be found in dumpsters


Other:
-Fixed the crafting for Akorus and Revolver Ultra (We do not have `Spring` as an item)
-Fixed Missing images from interaction menu for new items

minipunch (approved):

PR #1684, opened 2023-01-10
fix: Progress Bar Parts/Upgrades Dupe

Savage912-Official:
This will make it so the user can't open their inventory to dupe parts while installing upgrades and parts.

The set interaction busy code was already in the code for the parts/upgrades install to block the M menu when installing parts / upgrades so I made it to where if the set interaction is busy then it won't allow the open inventory keybind to be used.


minipunch:
Did u test to confirm it works?

minipunch (approved):

PR #1692, opened 2023-01-14
Disabled Blood Apartments

Urbaniized:


Urbaniized:
Building is unused and causes massive frame drop

minipunch (approved):

PR #1701, opened 2023-01-17
[Police] Added Police Impound

### LINKED ISSUE #789: police impound update
allow for impounding vehicle for up to X days

maxgestic:
This adds a police impound that allows BCSO and SASP officers to impound vehicles into police custody for up to 90 days. These vehicles can be retrieved by officers during this time for evidence gathering or other purposes. During this time the owner of the vehicle will not be able to retrieve their vehicles from any garage or the insurance. Police will also be able to release a vehicle early before it is meant to using `/releaseimpound [plate]`

Also Change the CouchDB views to the following:
getVehiclesForGarageMenu
```
function (doc) {
  emit(doc._id, [doc.plate, doc.make, doc.model, doc.impounded, doc.stored, doc.hash, doc.owner, doc.stats, doc.upgrades, doc.stored_location, doc.adv_impounded, doc.released_on]);
}
```

getVehiclesForMenuWithPlates
```
function (doc) {
  emit(doc._id, [doc.make, doc.model, doc.price, doc.stored, doc.stored_location, doc._id, doc.adv_impounded]);
}
```

Closes #789


minipunch:
Can you explain how one impounds a vehicle with this update for me plz

maxgestic:
> Can you explain how one impounds a vehicle with this update for me plz
> 
> Also FYI I am thinking 14 days to start, maybe 30 max

You just drive it to one of the impound locations and press E then enter a time you want to impound it for

minipunch:
Also with this it would be fire if we make a way for people to 'break in' and have some chance of getting their vehicle back (or failing)

minipunch:
Also it is preferable to move away from the views since it requires manual setup

There is probably a good enough way to do it using the normal DB API we have setup (i.e. `exports.essentialmode:getDocument` etc)

resources/usa_impound/client.lua
```diff
@@ -0,0 +1,196 @@
...
+		for i, isNearby in pairs(nearbyLocations) do
+			local info = impound_locations[i]
+			local dist = Vdist(GetEntityCoords(ped), info.x, info.y, info.z)
+			DrawMarker(27, vector3(info['x'], info['y'], info['z'] - 0.98), 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, vector3(4.0, 4.0, 3.0),  255, 0, 0, 150, false, true, 2, false, nil, nil, false)
+			if dist < 10 then
+				DrawText3D(info['x'], info['y'], info['z'], '[E] - Police Impound')
+			end
+			if IsControlJustPressed(0, 86) and dist < 2 then
+
+				TriggerServerEvent("usa_impound:checkCop")
```
minipunch, line 113: Can you use `pmc-callbacks` for this instead of events please

resources/usa_impound/server.lua
```diff
@@ -0,0 +1,165 @@
...
+RegisterServerEvent("usa_impound:impoundVeh")
+AddEventHandler("usa_impound:impoundVeh", function(plate,days)
+	local released_on = os.time() + days * 86400
+	TriggerEvent('es:exposeDBFunctions', function(db)
+		db.updateDocument("vehicles", plate, { adv_impounded = true, released_on = released_on }, function()
+	        db.getDocumentById("police-impound", "impoundedPlates", function(doc)
+	            if doc then
+	            	local plates = doc.plates
+	            	table.insert(plates,plate)
+	            	db.updateDocument("police-impound", "impoundedPlates", {plates = plates}, function()
```
minipunch, line 12: In Couch DBs docs it says that for best performance keep the doc size as little as possible

So instead of appending to one giant list in a single document we should create a new, small doc for each new impounded vehicle

resources/usa_impound/server.lua
```diff
@@ -0,0 +1,165 @@
...
+	            		end
+	            	end
+	            	db.updateDocument("police-impound", "impoundedPlates", {plates = plates}, function()
+	            		-- done
+	            	end)
+	            end
+	        end)
+	    end)
+		TriggerClientEvent("garage:vehicleStored", usource, vehicle)
+		local time_until = round((vehicle.released_on - os.time())/86400, 0)
```
minipunch, line 129:
```suggestion
		local time_until = math.floor((vehicle.released_on - os.time())/86400)
```

minipunch: Since it is built in

resources/usa_impound/server.lua
```diff
@@ -0,0 +1,165 @@
...
+		TriggerClientEvent("usa:notify", usource, "Vehicle ("..vehicle.plate..") retrieved from police impound! There where "..time_until.." days left! Make sure to book it back in!", "Vehicle ("..vehicle.plate..") retrieved from police impound! There where "..time_until.." days left! Make sure to book it back in!")
```
minipunch, line 130:
```suggestion
		TriggerClientEvent("usa:notify", usource, "Vehicle ("..vehicle.plate..") retrieved from police impound! There was "..time_until.." days left! Make sure to book it back in!", "Vehicle ("..vehicle.plate..") retrieved from police impound! There where "..time_until.." days left! Make sure to book it back in!")
```

resources/usa_garage/server.lua
```diff
@@ -136,6 +136,41 @@ AddEventHandler("garage:vehicleSelected", function(vehicle, business, playerCoor
 		else
 			TriggerClientEvent("usa:notify", usource, "~y~STATE IMPOUND: ~s~Your vehicle is impounded and can be retrieved for ~y~$"..IMPOUND_FEE..".00~s~!")
 		end
+	elseif vehicle.adv_impounded == true then
+		if (vehicle.released_on - os.time() > 0) then
+			local time_until = round((vehicle.released_on - os.time())/86400, 0)
```
minipunch, line 141:
```suggestion
			local time_until = math.floor((vehicle.released_on - os.time())/86400, 0)
```

built in and has no decimals

minipunch:
I impounded my vehicle but whenever I go to press E on the impound to open the menu it does not open and there aren't any errors thrown

I did update both views already

Does this not happen for you?

resources/usa_impound/server.lua
maxgestic: IDK since it only stores the plate it seems a bit idk weird to make a new sheet for just storing a place, I feel like for having an array of plates it should be fine keeping it in one doc no?

maxgestic: I mean plus having a separate sheet for each vehicle would also mean the database would be bigger overall cos if its all in one you are only storing the "_id" and "_rev" for one doc, but if you have it in separate docs then it be like multiple "_id" and "_rev"'s?

maxgestic:
> I impounded my vehicle but whenever I go to press E on the impound to open the menu it does not open and there aren't any errors thrown
> 
> I did update both views already
> 
> Does this not happen for you?

Works fine for me are you sure you where whitelisted and clocked on as a cop?
https://youtu.be/jBNtlvIfXPI

resources/usa_impound/client.lua
maxgestic: Changed this to use a more tradition back and forth

minipunch:
> > I impounded my vehicle but whenever I go to press E on the impound to open the menu it does not open and there aren't any errors thrown
> 
> > 
> 
> > I did update both views already
> 
> > 
> 
> > Does this not happen for you?
> 
> 
> 
> Works fine for me are you sure you where whitelisted and clocked on as a cop?
> 
> https://youtu.be/jBNtlvIfXPI

Yep I did /setjob sheriff and had rank 5, I can try again I guess..

resources/usa_impound/server.lua
minipunch: It's not about the DB being larger overall, it's about trying to keep individual doc sizes as small as possible

It's fine for this application since there most likely won't be all that many impounded plates to store

But in theory if this list grew huge then reading and writing to it would be less efficient than separate docs according to the performance pages of Couch DBs documentation

Mainly because updating requires reading the entire document from disk and re-writing it all back and disk I/O is the slowest part of a DB

For example if we just want to make some very minor change to a document, then having to read all of it and write all of it back will take much longer with a large document than a small one

See the first point, "file size": https://docs.couchdb.org/en/stable/maintenance/performance.html

resources/usa_impound/client.lua
minipunch: Still not using `pmc-callbacks` lol

AKA:

```
local isCop = TriggerServerCallback {
  eventName = "impound:copCheck",
  args = {}
}

minipunch (changes requested):
not using pmc callback still

minipunch:
I'll adjust it to use pmc callbacks

also prob gonna adjust how we store impounded car info to avoid having to use views and keep it isolated to a new DB

also prob gonna make it so u have to pay for every date it is stored at impound to retrieve it (like IRL lmao)

minipunch (approved):

PR #1707, opened 2023-01-18
Updated Boosting to v1.4

DatBoiWeep:
**Note for Mini**
* When you push this, make sure you run `db_cars.sql` so it can be heckin working.

**Changes**
* Group Boosting is added with up to 4 people you choose by entering the ID
     - Cash is divided by num of people in group (Ex: 4 in a group get 25% each)
     - More scoin is given to the main contract host
* XP values have been increased per class
* Lowered contract spawn rate (~50% Time reduction)
* Increased price range if the vehicle is damaged upon delivery
* Added new SQL DB to track boosted vehicles
     - Boosted vehicles will be tracked and ensured that the amount can not be resold for a high cost (Abusing = Punishment)
* Changed `usa_carshop` from `__resource.lua` to `fxmanifest.lua`
     - Needed to use `oxmysql`
* Updated Boosting Store
* Added debugs for me :)

**Fixes**
* Fixed `DABABY`, `Vapid Taxi`, `News Rumpo` vehicles not spawning by adding a condition upon adding to the garage. **Note:** These vehicles can not be sold at the Dealership since they are not in the dealership. (Dealership will post the message below)
image

* Fixed an issue with incorrect spelling/capitalization on hash models (Hummer `H2` for example)

**Known Bugs**
* BMWs can not be VIN scratched since they are separated from the `usa_carshop` list. As a result, they are temporarily disabled until I implement a function for them to be VIN scratched with info.

**Notes to Repo Contributors**
* If you add a vehicle that is **NOT** in the dealership, edit the `rahe-boosting/api/server.lua` file and locate the `restrictedVehModel` variable and simply add the vehicle spawn name in there.

**Images**
* image

**Changelog Update from Developer**
* 500 new vehicle spawn spots. This will bring a fresh feeling to your boosters, and won't allow them to remember the spots so easily anymore.
* Added a config option ejectFromInvalidVehicles (enabled by default). This will remove players from vehicles that aren't hacked yet (A/S class). This will add extra security in case you haven't done the suggested integration (that these vehicles shouldn't be lockpicked). 
* Made the hacking device vehicle unlocking function encrypted. This will help people with certain vehicle key scripts.
* Made ped spawning unencrypted. You can now change their behaviour like alertness, shoot rate, health etc.
* Moved vehicleAreaMaximumOffset config option from client to server config (to support continuing contracts after disconnect).
* Vehicle deleting moved to server-side (to avoid issues where the client lost entity ownership and the vehicle couldn't be deleted).
* Automatic version check on resource start-up. Will alert you directly in the console if a new version has been released.
* Disallowed using the hacking device inside the car for the 2nd time. Previously, some people accidentally used this again and restarted their GPS hacking progress.
* All pictures in tablet/nui/img now loaded by default (instead of having to add your pictures file names to fxmanifest).
* Ability to disable the commands via a config option (instead of having to delete / comment them out).
* Renamed rahe-hackingdevice build folder from build to built-source (to avoid complaints about 'UI not showing' because 'build' folder didn't get pushed to their live server).
* Allow VIN scratching to be turned off from the config.
* Multiple security updates.
* Performance updates (more caching in places).
* UI improvements (smoother animations, notifications won't move the UI anymore etc).


minipunch:
This ready for review? @DatBoiWeep

DatBoiWeep:
> This still ready for review? @DatBoiWeep 

Si

minipunch:
This looks like a lot of changes

I'm scared

Hopefully u tested it lol

minipunch (approved):

PR #1733, opened 2023-01-25
update: Ability to break out of cuffs

### LINKED ISSUE #1727: allow players to 'fight back' when being cuffed
like some mini game that appears where it gives them a chance to break free

ideally it would be like COP vs. CRIM and whoever wins the mini game wins the cuffing battle

but if not that then at least letting the crim do some mini game for a chance to prevent the cuff would work for now

Savage912-Official:
Your cuff counts starts out at zero, Each time you get cuffed it goes up by 1, If you get cuffed more than 3 times on the 4th time you will be auto cuffed with no minigame. If you broke out of cuffs your count will reset after 30 minutes, but also if you are cuffed your cuff count is also reset to 0.

closes #1727


resources/usa_cuffs/cl_cuff.lua
```diff
@@ -286,26 +290,68 @@ AddEventHandler("cuff:Handcuff", function(arrestingPlayerId, x, y, z, playerHead
...
+						TriggerEvent("cuff:playSuspectAnim", x, y, z, playerHeading)
+
+						local success = lib.skillCheck({ {areaSize = 50, speedMultiplier = 1.5}, {areaSize = 40, speedMultiplier = 2}, {areaSize = 30, speedMultiplier = 2.7} })
+						if success then
+							count = count + 1
+							isCuffed = false
+							isHardcuffed = false
+							TriggerEvent('usa:showHelp', false, 'You have ~g~Escaped!')
+
+							Wait(1800000)
```
minipunch, line 335: You should use `GetGameTimer()` here instead since it is more reliable than using `Wait` since wait depends on the player's FPS whereas the game timer doesn't

See examples in other scripts

minipunch (changes requested):
see comment

resources/usa_cuffs/cl_cuff.lua
```diff
@@ -332,8 +332,11 @@ AddEventHandler("cuff:Handcuff", function(arrestingPlayerId, x, y, z, playerHead
 							isHardcuffed = false
 							TriggerEvent('usa:showHelp', false, 'You have ~g~Escaped!')
 
-							Wait(1800000)
-							count = 0
+							local beginTime = GetGameTimer()
+							while GetGameTimer() - beginTime < 1800000 do
+								Citizen.Wait(0)
+								count = 0
```
minipunch, line 338: This doesn't need to be in the `while` loop, just move it to just below it

minipunch (changes requested):

minipunch (approved):

minipunch:
nice!!

PR #1741, opened 2023-01-26
Updates to Racing

Urbaniized:
Moved Race start time to 5PM PST - Allows for EU and EST people to come
Increased Amount of Races
Adjusted chances of A and B class races to be more frequent


minipunch:
How about 6 PM PST

Urbaniized:
6 PM PST IT IS

PR #1754, opened 2023-01-31
Added Garage location

Rezitoo:
Added by the beach


minipunch (changes requested):
looks like you need to update your branch with the latest from `public-server`

this has way more changes than just a garage addition

Rezitoo:
Okay I fix

PR #1758, opened 2023-02-01
Police Lambo Updates

Urbaniized:
Removed the Lambo from the PD spawn menu and prison menu. Lamborghini can only be taken out with `/spawn` and only by command. SGT+ in BCSO or Lt+ in SASP.


Urbaniized:
Forgot to say, removed it from the PD menu to avoid abuse so not everyone who clocks in can just take one out.

minipunch (approved):

PR #1795, opened 2023-02-13
Confirmation box before purchasing vehicle

DatBoiWeep:
Because we probably needed this a few years ago

image
image


minipunch:
Does this work if the person buys > 1 vehicle in any given day tho ???

minipunch (approved):

PR #1816, opened 2023-02-17
update: Minimap Veh/Foot Settings

Savage912-Official:
You can now toggle the vehicle & foot minimap on/off meaning if you have foot on but vehicle off it won't show in the vehicle etc...


resources/usa_hud/client.lua
```diff
@@ -251,14 +252,25 @@ function IsBeltVehicle(vehicle)
...
-		exports.globals:notify("Minimap enabled")
+		exports.globals:notify("Foot Minimap enabled")
 	else
-		exports.globals:notify("Minimap disabled")
+		exports.globals:notify("Foot Minimap disabled")
+	end
+end)
+
+RegisterNetEvent("usa_hud:ToggleMinimapVeh") 
+AddEventHandler("usa_hud:ToggleMinimapVeh", function(status)
```
minipunch, line 267: to reduce copy and pasted code which can causes problems and clutter later:

just pass in a new argument to the more generic version, so smth like: `ToggleMinimap(status, type)`

then we don't need a separate event

resources/usa_hud/server.lua
```diff
@@ -28,32 +28,53 @@ end, {
 RegisterServerCallback {
     eventName = "usa_hud:GetMapSettings",
     eventCallback = function(source)
-        local enabled = false
+        local footenabled = false
```
minipunch, line 31: Would be ideal to move the defaults into a common config file shared between server and client so we don't have to worry about keeping stuff synced in diff files buried in code

Like move the current `Config` object in the client script to a `shared` `config.lua` script

minipunch (changes requested):

resources/usa_hud/server.lua
```diff
@@ -28,32 +28,55 @@ end, {
...
         end)
-    end)
+    else
+        TriggerClientEvent("usa:notify", source, "Invalid Arguments Provided!")
+    end
 end, {
-	help = "Enable or disable minimap"
+	help = "Enable or disable minimap",
+    params = {
+        { name = "Veh or Foot", help = "Toggle Vehicle & Foot Minimap On/Off" }
```
minipunch, line 78:
```suggestion
        { name = "type", help = "either 'veh' or 'foot'" }
```

resources/usa_hud/server.lua
```diff
@@ -28,32 +28,55 @@ end, {
...
+                    TriggerClientEvent("usa_hud:ToggleMinimap", src, veh, doc.vehenabled)
+                else
+                    db.createDocumentWithId("minimap-settings", {vehenabled = true}, id, function() end)
+                    TriggerClientEvent("usa_hud:ToggleMinimap", src, veh, true)
+                end
+            end)
         end)
-    end)
+    else
+        TriggerClientEvent("usa:notify", source, "Invalid Arguments Provided!")
```
minipunch, line 73:
```suggestion
        TriggerClientEvent("usa:notify", src, "Invalid Arguments Provided!")
```

resources/usa_hud/server.lua
```diff
@@ -28,32 +28,55 @@ end, {
...
-                db.updateDocument("minimap-settings", id, {enabled = doc.enabled}, function() end)
-                TriggerClientEvent("usa_hud:ToggleMinimap", src, doc.enabled)
-            else
-                db.createDocumentWithId("minimap-settings", {enabled = true}, id, function() end)
-                TriggerClientEvent("usa_hud:ToggleMinimap", src, true)
-            end
+
+    if args[2] == "foot" then
+        TriggerEvent('es:exposeDBFunctions', function(db)
+            db.getDocumentById("minimap-settings", id, function(doc)
```
minipunch, line 48: while we're here can you update this db stuff to use the new `exports.essentialmode...` format please since it is cleaner

resources/usa_hud/server.lua
```diff
@@ -28,32 +28,55 @@ end, {
...
+                    db.updateDocument("minimap-settings", id, {footenabled = doc.footenabled}, function() end)
+                    TriggerClientEvent("usa_hud:ToggleMinimap", src, foot, doc.footenabled)
+                else
+                    db.createDocumentWithId("minimap-settings", {footenabled = true}, id, function() end)
+                    TriggerClientEvent("usa_hud:ToggleMinimap", src, foot, true)
+                end
+            end)
+        end)
+    elseif args[2] == "veh" then
+        TriggerEvent('es:exposeDBFunctions', function(db)
```
minipunch, line 60: don't forget here too

resources/usa_hud/client.lua
```diff
@@ -252,19 +248,29 @@ function IsBeltVehicle(vehicle)
...
+	      if status then
+		  exports.globals:notify("Vehicle Minimap enabled")
+	      else
+		  exports.globals:notify("Vehicle Minimap disabled")
+	      end
+       end
 end)
 
 Citizen.CreateThread(function()
 	Config.MapEnabled = TriggerServerCallback {
```
minipunch, line 272: Looks like u forgot to update this for when they load in

resources/usa_hud/server.lua
Savage912-Official: Alright I’ll do it in a bit

resources/usa_hud/client.lua
```diff
@@ -252,18 +248,29 @@ function IsBeltVehicle(vehicle)
...
+		  exports.globals:notify("Vehicle Minimap enabled")
+	    else
+		  exports.globals:notify("Vehicle Minimap disabled")
+	    end
+    end
 end)
 
 Citizen.CreateThread(function()
-	Config.MapEnabled = TriggerServerCallback { 
+    enabled = Config.VehMapEnabled and Config.FootMapEnanled
```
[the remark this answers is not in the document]
minipunch: 1) "FootMapEnanled"
2) this line is overwritten by the line just after it ??

resources/usa_hud/server.lua
```diff
@@ -28,32 +28,48 @@ end, {
...
-    end)
+    local doc = exports.essentialmode:getDocument("minimap-settings", id)
+
+    if args[2] == "foot" then
+        if doc then
+            doc.footenabled = not doc.footenabled
+            exports.essentialmode:updateDocument("minimap-settings", id, {footenabled = doc.footenabled}, true)
+            TriggerClientEvent("usa_hud:ToggleMinimap", src, "foot", doc.footenabled)
+        else
+            exports.essentialmode:createDocumentWithId("minimap-settings", id, {footenabled = true})
```
minipunch, line 53: No need for this separate `else` branch just to make the doc if it doesn't exist

`updateDocument` does that already since u passed `true` as the last argument already

resources/usa_hud/server.lua
```diff
@@ -28,32 +28,48 @@ end, {
...
+            TriggerClientEvent("usa_hud:ToggleMinimap", src, "foot", true)
+        end
+    elseif args[2] == "veh" then
+        if doc then
+            doc.vehenabled = not doc.vehenabled
+            exports.essentialmode:updateDocument("minimap-settings", id, {vehenabled = doc.vehenabled}, true)
+            TriggerClientEvent("usa_hud:ToggleMinimap", src, "veh", doc.vehenabled)
+        else
+            exports.essentialmode:createDocumentWithId("minimap-settings", id, {vehenabled = true})
```
minipunch, line 62: same here

resources/usa_hud/server.lua
```diff
@@ -28,32 +28,42 @@ end, {
...
-            else
-                db.createDocumentWithId("minimap-settings", {enabled = true}, id, function() end)
-                TriggerClientEvent("usa_hud:ToggleMinimap", src, true)
-            end
-        end)
-    end)
+    local doc = exports.essentialmode:getDocument("minimap-settings", id)
+
+    if args[2] == "foot" then
+        if doc then
```
minipunch, line 48: don't really need to check for whether `doc` exists either lol

Savage912-Official: 😂

resources/usa_hud/client.lua
```diff
@@ -108,19 +104,19 @@ Citizen.CreateThread(function()
 				else
 					DrawTxt(0.663, 1.455, 1.0, 1.0, 0.40, hud.time, 255, 255, 255, 255)
 				end
-				if Config.MapEnabled then
+				if Config.VehMapEnabled and enabled then
```
minipunch, line 107: where does this `enabled` variable come from ??

minipunch:
```suggestion
				if Config.VehMapEnabled and hud.enabled then
```

right ?

minipunch:
had some issues

for now I just did it this way:

[a8f9fe3](https://github.com/minipunch/usarealismrp-fxserver/commit/a8f9fe3780883230d792815090e40cc8a5b78937)

PR #1831, opened 2023-02-21
Mech Location Update

Urbaniized:
Removed old Benny's stuff
Added Pressure Washer to new Benny's
Added mech sign-in for Alpha Auto Service Center (Rezito)


minipunch (approved):
how much we betting on a missed comma

PR #1832, opened 2023-02-21
Added AASC Mechanic stuffs

Rezitoo:
Honestly if there are way more changes or things are out of date I am truly lost. These updates are based on the absolute most updated versions of the files and I cannot for the life of me tell if there is a problem


Rezitoo:
Also the interior is ready just need to add it to ghub

Rezitoo:
Never mind I want death

Issue #1224, opened 2022-09-07
way to edit a property's price


minipunch:
for now prob a new command for staff would suffice

something like: `/editproperty [name] <new info...>`

minipunch:
#1838 will close this

PR #1838, opened 2023-02-21
[Properties] Property System Changes

maxgestic:
- Added ability to transfer property to another player
- Changed the way Mod+ add properties to a much nicer experience (https://youtu.be/sTQR7bQZLNY)
- Added /editproperty for Mod+ for easy in game editing of a properties locations, name and prices (https://www.youtube.com/watch?v=hSFgVrKk230)
- Added checks to prevent properties with the same name for both adding and editing
- Properties will now take money from the bank every 30 days instead of the money inside of the property
- There is a new option to leave property in property menu that if unchecked will end your tenancy at the next due date
- Co-Owners will be able to see if the property will continue to be owned or not
- All current properties will be set to continue being owned so if you do not want to keep your property then go and de-select the option
- Added /addpropertydoor command Mod+ to add lockable doors to a property ( https://www.youtube.com/watch?v=fQiAdSBTHkI)
- Added /removepropertydoor command Mod+ to remove lockable doors from a property (Only use in emergency, is a bit finicky)
- Property Doors can be locked/unlocked by both the Owner of a property as well as any Co-Owners
- Added Police Battering Ram that can force open any property door (though you might have to hit a couple of times) to use just walk up to property door and "Use" the battering ram via the inventory, this will play an auditable sound to any players in the vicinity 
- Property Doors will be unlocked if no-one owns the property but lock automatically when someone purchases the property and while property is owned will always be locked by default


resources/usa-properties-og/server.lua
```diff
@@ -679,6 +713,69 @@ AddEventHandler("properties:addCoOwner", function(property_name, id)
...
+AddEventHandler("properties:changeOwner", function(property_name, id)
+  if source ~= id then
+    if GetPlayerName(id) then
+      -- create person --
+      local person = exports["usa-characters"]:GetCharacter(id)
+      if GetNumberOfOwnedProperties(person.get("_id")) < MAX_NUM_OF_PROPERTIES_SINGLE_PERSON then
+        if PROPERTIES[property_name].lastTransfer == nil then
+          PROPERTIES[property_name].lastTransfer = 0
+        end
+        if os.time() - tonumber(PROPERTIES[property_name].lastTransfer) > 172800 then
```
minipunch, line 726: Can u make this `172800` number into something a bit more meaningful like a named 'const' variable defined at the top plz

resources/usa-properties-og/server.lua
```diff
@@ -679,6 +713,69 @@ AddEventHandler("properties:addCoOwner", function(property_name, id)
   end
 end)
 
+RegisterServerEvent("properties:changeOwner")
+AddEventHandler("properties:changeOwner", function(property_name, id)
```
minipunch, line 717: No security check so clients can't arbitrarily execute this to change ownership of houses without permission?

minipunch: Same for the other critical events

maxgestic: Fixed

resources/usa-properties-og/server.lua
```diff
@@ -13,6 +13,8 @@ local NEARBY_DISTANCE = 500
 
 local CLIENT_UPDATE_INTERVAL = 10
 
+local PROPERTY_TRANSFER_TIMEOUT = 48 -- hours
```
minipunch, line 16:
```suggestion
local PROPERTY_TRANSFER_TIMEOUT_HOURS = 48
```

resources/usa-properties-og/server.lua
```diff
@@ -854,34 +974,41 @@ function Evict_Owners()
...
+            if ownerPlayer.bank >= info.fee.price and not info.will_leave then
+              print("** Money from bank of (" .. ownerPlayer.name.first .. " " .. ownerPlayer.name.last .. ") was used to pay for a month of rent. **")
+              ownerPlayer.bank = ownerPlayer.bank - info.fee.price
+              exports["essentialmode"]:updateDocument("characters", PROPERTIES[name].owner.identifier, ownerPlayer, true)
+              PROPERTIES[name].fee.paid_time = os.time()
+              PROPERTIES[name].fee.paid = true
+              local final_time = nil
+              local today = os.date("*t", os.time())
+              PROPERTIES[name].fee.due_days = HOUSE_PAY_PERIOD_DAYS
+              final_time = {day = today.day, month = today.month + 1, year = today.year}
```
minipunch, line 988: just realized this is kinda sketch cuz what happens if `today.month` is `12` (aka december)

then we add 1 to make it `13`?? it should wrap around back to like 1 probably

minipunch, line 988:
```suggestion
              final_time = {day = today.day, month = (today.month + 1) % 12, year = today.year}
```

should be more what we want

minipunch (approved):

minipunch:
this means that people who are negative in the bank are going to lose their properties at restart right lol

minipunch:
for github: this will close #1224

maxgestic:
> 

Yes, this is true

maxgestic:
> this means that people who are negative in the bank are going to lose their properties at restart right lol
> 
> i think before we push this we prob should make an announcement so people are aware or smth lol

Yea this is true and good idea, ill make an accountment

maxgestic:
OK all good!

PR #1839, opened 2023-02-22
removed fight club blip 

Zeeroqt:
as title says


minipunch:
can u do it in the same PR lol

PR #1842, opened 2023-02-22
update: Lottery Updates

Savage912-Official:
**Changes Made:**

- Cleaned up the menu a bit
- Added the comma_value for ticket price
- Added Previous Winner & Amount
- Fixed a bug where when the prize is claimed it doesn't remove the users from the database resulting in them not being able to enter the new lottery.
- Updated the SQL for the new changes

**You may wanna manually insert the 2 new tables so it doesn't mess anything up but I did add them to the sql as well.**

`previoustotal` int(4) DEFAULT NULL,
`previouswinner` varchar(255) CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci DEFAULT NULL

Previews:
https://imgur.com/a/5yPHNp7


resources/usa_lottery/server/sv_lottery.lua
```diff
@@ -23,10 +23,12 @@ RegisterNetEvent("usa_lottery:choosewinner", function()
     local result = MySQL.query.await('SELECT * FROM lotterytotal where lotto = ?', {check})
     local govAmount = Config.Gov_Percentage * result[1].total
     local restAmount = result[1].total - govAmount
+    local previoustotal = result[1].total - govAmount
```
minipunch, line 26: shouldn't really need to define a new variable when the same one is right above it lol

minipunch (approved):

PR #1844, opened 2023-02-23
New chatbox for usa

Zeeroqt:
- New UI 
- New Font
- Cool fading in transition when opening chat box.
- Preview: https://streamable.com/5yk3m8

### DIFF
--- a/resources/[system]/chat/html/index.css
+++ b/resources/[system]/chat/html/index.css
@@ -1,123 +1,159 @@
-.color-0{color: #ffffff;}
-.color-1{color: #ff4444;}
-.color-2{color: #99cc00;}
-.color-3{color: #ffbb33;}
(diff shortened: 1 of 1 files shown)


minipunch:
bruh it gotta have chat command suggestions at minimum

minipunch:
would be ideal if we could make it optional so people can use either (or 'opt in' to try it)

Zeeroqt:
> would be ideal if we could make it optional so people can use either (or 'opt in' to try it)

huh

minipunch:
> > would be ideal if we could make it optional so people can use either (or 'opt in' to try it)
> 
> huh

u know

just in case some people hate it and want the old one

Zeeroqt:
oh um i mean i can see what we can do or get a vote

PR #1850, opened 2023-02-23
Benny's, LaMesa & Atomic ATM's

Urbaniized:
Added ATM's for the MLOs


minipunch (approved):

minipunch (changes requested):
Not sure editing gabz' maps is a good idea

Probably gonna end up losing them in the next update

Urbaniized:
It's just the YTYP which is just interior props, doesn't change anything about the model it self. Gabz leaves it un-encrypted so people can edit it

minipunch:
> It's just the YTYP which is just interior props, doesn't change anything about the model it self. Gabz leaves it un-encrypted so people can edit it

Still might end up losing it in an update since I have to drag and drop everything over lol

Urbaniized:
I keep backups of them lol, it's not hard to add the extra props. I can see if I can get the additional props on a new file that is not gabz

Urbaniized:
Oh I also updated some of the graffiti but that's not hard either

minipunch (approved):

PR #1863, opened 2023-02-27
update: Skill Checks Inputs

Savage912-Official:
With the new ox_lib updates you can now choose which inputs you want to use.


minipunch:
What's the status on this?

Savage912-Official:
Need the latest ox_lib problem is it breaks the wheel fitment script and made a ticket in the owners discord and still haven’t gotten a response.

PR #1879, opened 2023-03-03
update: Repair Tools

### LINKED ISSUE #1862: Non Mechanic Repair Kits
https://discord.com/channels/297566457468944384/363082636505382912/1079095171507048448

Savage912-Official:
Repair tools are a 1 time use if the repair is successful if it fails there is a chance the repair tools breaks. You don't have to be on duty to use this item.

Currently the only way to obtain the repair tools is by crafting them as a level 3 mechanic.

closes #1862


resources/usa_mechanicjob/mechanic_cl.lua
```diff
@@ -668,6 +668,48 @@ AddEventHandler("mechanic:usedPart", function(part)
 	TriggerServerEvent("mechanic:usedPart", part, vehPlate)
 end)
 
+RegisterNetEvent("mechanic:repairtools")
+AddEventHandler("mechanic:repairtools", function()
+	local veh = exports.globals:getClosestVehicle(3.0)
```
minipunch, line 673: Did u just copy and paste this code from some other function or event?

If so can we move this into some shared common function they can both use instead of copy n paste

minipunch: Oh it looks like you added a skill check thing which is slightly different, and new I think

Maybe u still can add like an optional param to do the skill check if that is the only difference in some shared function

Savage912-Official: Yeah I can take a look and add it to the mechanic helper.repair vehicle function

resources/usa_mechanicjob/classes/MechanicHelper_cl.lua
```diff
@@ -145,57 +145,79 @@ MechanicHelper.getClosestVehicle = function(maxRange)
...
-                if not IsVehicleDriveable(veh, true) then -- damaged and red
-                    SetVehicleUndriveable(veh, false)
-                    SetVehicleEngineHealth(veh, 500.0)
-                else -- Damaged but not red, so prob orange
-                    SetVehicleEngineHealth(veh, 800.0)
+        if not IsPedInVehicle(PlayerPedId(), veh, false) then
+            SetVehicleDoorOpen(veh, 4, false, false)
+            TriggerEvent("dpemotes:command", 'e', GetPlayerServerId(ped), {"mechanic"})
+
+            if type == "advanced" then
```
minipunch, line 157: Why name this type `"advanced"` ?

Savage912-Official: True ill rename

resources/usa_mechanicjob/mechanic_cl.lua
```diff
@@ -668,6 +668,28 @@ AddEventHandler("mechanic:usedPart", function(part)
 	TriggerServerEvent("mechanic:usedPart", part, vehPlate)
 end)
 
+RegisterNetEvent("mechanic:repairtools")
+AddEventHandler("mechanic:repairtools", function()
+	if not isRepairing then
+		isRepairing = true
+		local veh = MechanicHelper.getClosestVehicle(5)
+		if veh then
+			MechanicHelper.repairVehicle(veh, "tools", repairCount, function(success)
```
minipunch, line 677: `repairCount` is missing

this is why copy n pasting code is bad lol

unless u meant to remove that param from the event..

but either way this event seems like a good candidate for just adding another param instead of copy n pasting into a new event

minipunch: It looks like u lucked out though and here it still works.. but definitely not ideal lol

minipunch:
Nice

I think it would be cool if we make both repair types use that skill check game

but then make it so the more success repairs u have under ur belt the easier the game gets

minipunch (approved):

PR #1883, opened 2023-03-06
Added charges to /jail

Rezitoo:
Made some charges felonies

### DIFF
--- a/resources/[system]/globals/sv_global.lua
+++ b/resources/[system]/globals/sv_global.lua
@@ -266,7 +266,7 @@ end
 
 function hasFelonyOnRecord(src)
 	local felonyChargeNumbers = {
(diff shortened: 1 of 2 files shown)


minipunch:
Don't think 23109(b) AKA speed contest is usually charged as a felony, it's generally a misdemeanor here in CA

Rezitoo:
Yeah when I looked it up (Street Racing) is a felony in CA which is why I added (b) to the law book. Unfortunately the only way to revoke a license when arresting someone for illegal street racing is to make it a felony

minipunch:
I think it would only be charged as a felony if u crashed / hurt someone

And really you don't just lose your license for a 'street racing' ticket either

The DMV has a 'point' system though and if you have enough points basically or smth like that you can have it suspended

I'd like to get that system added

Cause it also usually raises insurance prices and crims will LOVE that 😂

minipunch (changes requested):
see comment

Rezitoo:
Alright that works, will delete this and wait to give points for it lmaooo

PR #1892, opened 2023-03-07
update: Joint Updates

### LINKED ISSUE #1891: Joint Updates
https://usarrp.gg/forums/topic/1856-high-effect/

I’ll change it tomorrow

Savage912-Official:
Changes Made:

- Removed The sv_config & cl_config and replaced with a shared config file as there was no need for both.

- Adjusted the Weed effect time from 10 minutes to 3 (There was a suggestion for it to be lowered as it can cause vision issues with graphics mods)

- Changed the joint animation flag to allow movement while smoking.

closes #1891


minipunch:
Can u make a separate join item (like a new strain or smth) that has the reduced vision effect duration instead plz ?

Cuz I think some people (including myself) actually like the effect since we don't use gfx mods

Savage912-Official:
Sure

minipunch (approved):

PR #1899, opened 2023-03-08
update: Mechanic Tools Skillcheck

### LINKED ISSUE #1898: Mechanic Tools Skillcheck
Make the Mechanic Tools use skillcheck just like the repair kit and adjust the difficulty based on mechanic level.

Savage912-Official:
Added a skillcheck for the Mechanic Tools item rather than just a progress bar and made the skillcheck difficulty based on mechanic rank.

closes #1898


resources/usa_mechanicjob/classes/MechanicHelper_cl.lua
```diff
@@ -153,37 +152,24 @@ MechanicHelper.useMechanicTools = function(veh, repairCount, cb)
...
-                canCancel = true,
-                disable = {
-                    car = true,
-                    move = true,
-                    combat = true,
-                }
-            }) then 
-                local failChance = 0.5 - (0.005 * repairCount) -- larger successful repair count = smaller fail chance
-
-                if math.random() > failChance then
```
minipunch, line 177: I'll have to try this myself but if the skill checks are too easy we should re add this part, so they can just retry if it fails

Savage912-Official: Yeah we can re add it back just lmk what you think

minipunch: adjusted it a bit

minipunch (approved):

PR #1901, opened 2023-03-08
hardware store scuff

Zeeroqt:
Fix crowbar duplicate 
removed drill bc im making it where u can craft it and u can buy it already at a npc


minipunch:
stale

PR #1912, opened 2023-03-11
New usable inventory slot

Zeeroqt:
- Players are now able to use the 5th slot in their inventory with there number 5 key.
- Preview : https://streamable.com/hee7uc


minipunch (approved):

minipunch:
Nice

PR #1925, opened 2023-03-14
update: Disabled Breaking Cuffs While Dead

Savage912-Official:
Disabled the ability the break cuffs while dead and cleaned up a bit of code rather than repeating the cuff code for each statement I just made it a function to avoid repetitive code.


minipunch (approved):

minipunch:
and hopefully u tested it right

PR #1938, opened 2023-03-19
Little MDT change

Rezitoo:
Changed the wording for no weapon serial number result in mdt to accommodate for crafted weapons


minipunch:
How does this "accommodate for crafted weapons" exactly ?

PR #1974, opened 2023-03-24
Updated Aircraft shop

Zeeroqt:
- Renting an aircraft now takes it from your bank or cash
- Claiming insurance on a aircraft now also gets removed from your bank or cash
- Buying an aircraft now can take it from your cash or bank


minipunch (approved):
nice

PR #1980, opened 2023-03-27
Slash Tires

Rezitoo:
Added the ability for players to use Sharp Objects to Slash vehicle tires, but watch out there is a chance 911 gets a call!

Uses third eye to slash tire
Chance of missing
911 Call


Rezitoo:
https://i.imgur.com/kuC6OGT.png

Savage912-Official:
Nice!

minipunch:
Well first attempt didn't go so well...

minipunch:
Not sure if this is cause of your script though cause it just happened again without event trying.. all I did was open F8

minipunch:
Don't think u have to worry about that stuff, think that is a FiveM issue

but you did forget to start `slashtires` in `server.cfg`

minipunch:
Also like you mentioned this removes the `TAB` to put away weapon hotkey

We should indeed fix that so that still works

What was the reason for that here??

resources/interaction-menu/cl_thirdEye.lua
```diff
@@ -1,49 +1,21 @@
...
+	["wheel_rm2"] = 47,
+	["wheel_rm3"] = 48
+}
+
 function onInteract(targetData,itemData)
     print(targetData.name,targetData.label) --> my_buffalo_target   Buffalo
     print(itemData.name,itemData.label)     --> lock_door           Lock Door
 end
 
-function onATMInteract(targetData, itemData)
```
minipunch, line 6: Looks like you merged incorrectly here

We need these functions (this is stuff added not long ago)

resources/interaction-menu/cl_thirdEye.lua
```diff
@@ -80,6 +52,10 @@ function onVehicleOptionSelect(a, buttonInfo, hitHandle)
         end
     elseif buttonInfo.label == "Stickers" then
         TriggerEvent("rcore_stickers:open")
+    elseif buttonInfo.label == "Slash Tire" then
+        for name, index in pairs(bones) do
```
minipunch, line 56: Where is `bones` defined here?

resources/interaction-menu/cl_thirdEye.lua
```diff
@@ -449,45 +429,8 @@ function addMechanicVehicleOptions()
     })
 end
 
-function addCivModelOptions()
```
minipunch, line 452: again lost stuff

resources/interaction-menu/inventory/hotkeys/cl_hotkeys.lua
```diff
@@ -55,7 +55,7 @@ Citizen.CreateThread(function()
                         vehiclePlate = exports.globals:trim(vehiclePlate)
                     end
                     -- cycle through slots keys --
-                    if IsDisabledControlJustPressed(24, KEYS.TAB) then
+                    if IsDisabledControlJustPressed(24, KEYS.CTRL) then
```
minipunch, line 58: Why change this ?

minipunch (changes requested):

Rezitoo:
> Also like you mentioned this removes the `TAB` to put away weapon hotkey
> 
> 
> 
> We should indeed fix that so that still works
> 
> 
> 
> What was the reason for that here??

So the script requires the weapon to be in hand to actually slash the tire and with pressing tab to open the third eye to begin to slash the tire would in turn put the weapon away making slashing the tire impossible

resources/interaction-menu/cl_thirdEye.lua
Rezitoo: Bones are defined at the beginning of the script https://i.imgur.com/zjr8Wlr.png

resources/interaction-menu/cl_thirdEye.lua
Rezitoo: Will get it added

resources/interaction-menu/cl_thirdEye.lua
Rezitoo: Will get it added

resources/interaction-menu/inventory/hotkeys/cl_hotkeys.lua
Rezitoo: This was my attempt to change the key to put away the weapon, Is what enabled the script to work with the third eye in the first place

resources/interaction-menu/cl_thirdEye.lua
DatBoiWeep: should be back now

resources/interaction-menu/cl_thirdEye.lua
DatBoiWeep: should be back now

Savage912-Official:
> > > Also like you mentioned this removes the `TAB` to put away weapon hotkey
> > > We should indeed fix that so that still works
> > > What was the reason for that here??
> > 
> > 
> > So the script requires the weapon to be in hand to actually slash the tire and with pressing tab to open the third eye to begin to slash the tire would in turn put the weapon away making slashing the tire impossible
> 
> To avoid changing the third eye hotkey we should just give the player a knife (if they have one in their inv) before we call the slash tire event and it ought to work then right

We

> > > Also like you mentioned this removes the `TAB` to put away weapon hotkey
> > > We should indeed fix that so that still works
> > > What was the reason for that here??
> > 
> > 
> > So the script requires the weapon to be in hand to actually slash the tire and with pressing tab to open the third eye to begin to slash the tire would in turn put the weapon away making slashing the tire impossible
> 
> To avoid changing the third eye hotkey we should just give the player a knife (if they have one in their inv) before we call the slash tire event and it ought to work then right

Yeah that would work

minipunch (approved):

PR #1991, opened 2023-03-30
Door Tuning Fix

Urbaniized:
The trains were overriding the `gta5.meta` in the `mapdata` folder. Fixed the conflict and tested using the latest server resources.


resources/[trains]/max_trains/gta5.meta
```diff
@@ -1135,7 +1135,7 @@
 	  <fileType>OBJ_COVER_TUNING_FILE</fileType>
 	</Item>
 	<Item>
-	  <filename>platform:/levels/gta5/doortuning</filename>
```
Comment (minipunch) [inline resources/[trains]/max_trains/gta5.meta:1138]: @maxgestic Do we even need this door tuning line or can we remove this entry all together so we don't have to worry about it overwriting gabz' ?

Comment (Urbaniized) [reply resources/[trains]/max_trains/gta5.meta:1138]: I tried removing it, still did not work. We need to have it defined in this `gta5.meta`

Comment (minipunch) [reply resources/[trains]/max_trains/gta5.meta:1138]: I guess we would probably need to remove the whole meta file if possible

Comment (Urbaniized) [reply resources/[trains]/max_trains/gta5.meta:1138]: There are some things in there defined so the trains will work.

The Gabz meta file should not affect this since this one starts first. This is probably the easiest way to combine both the files for now

PR #2025, opened 2023-04-13
update: Weed Plant Whitelisted Locations

Savage912-Official:
Added whitelisted locations for places that are "on the road" but not actually on the road. Also added 2 new process locations as requested, and adjusted the range just a bit.


resources/usa_cultivation/cl_cultivation.lua
```diff
@@ -22,6 +22,23 @@ local KEYS = {
...
+	{
+		coords = vector3(-8.560081, -2519.72, -9.939499),
+		distance = 20
+	},
+}
+
+local function isAtWhitelistedLocation()
+	for i = 1, #Whitelisted_Plant_Locations do
+        local x, y, z = table.unpack(Whitelisted_Plant_Locations[i].coords)
+        if Vdist(GetEntityCoords(PlayerPedId()), x, y, z) < Whitelisted_Plant_Locations[i].distance then
```
minipunch, line 35:
```suggestion
        if Vdist(GetEntityCoords(PlayerPedId()), x, y, z) < Whitelisted_Plant_Locations[i].distance then
        if #(GetEntityCoords(PlayerPedId()) - Whitelisted_Plant_Locations[i].coords) < Whitelisted_Plant_Locations[i].distance then
```

minipunch: More concise (and I believe more optimized)

resources/usa_cultivation/cl_cultivation.lua
```diff
@@ -22,6 +22,23 @@ local KEYS = {
...
+local Whitelisted_Plant_Locations = {
+	{
+		coords = vector3(-8.560081, -2519.72, -9.939499),
+		distance = 20
+	},
+}
+
+local function isAtWhitelistedLocation()
+	for i = 1, #Whitelisted_Plant_Locations do
+        local x, y, z = table.unpack(Whitelisted_Plant_Locations[i].coords)
```
minipunch, line 34:
```suggestion
```
copy pasta much ?

Savage912-Official: Prob this wasn’t the original way I set it up I had it a different way but this is the one that got submitted.

resources/usa_cultivation/cl_cultivation.lua
```diff
@@ -22,6 +22,23 @@ local KEYS = {
...
+        if #(GetEntityCoords(PlayerPedId()) - Whitelisted_Plant_Locations[i].coords) < Whitelisted_Plant_Locations[i].distance then
```
minipunch, line 34:
```suggestion
        if #(playerCoords - Whitelisted_Plant_Locations[i].coords) < Whitelisted_Plant_Locations[i].distance then
```

since calling natives is CPU intensive we should try to minimize the amount we have to use them

here that means instead of calling `GetEntityCoords` every time in this `for` loop, let's just call it once before the loop and store it in a variable that we can use to check each item against so we only call it one time

resources/usa_cultivation/cl_cultivation.lua
```diff
@@ -22,6 +22,23 @@ local KEYS = {
...
 local POLL_INTERVAL_SECONDS = 5
 
+local Whitelisted_Plant_Locations = {
+	{
+		coords = vector3(-8.560081, -2519.72, -9.939499),
+		distance = 20
+	},
+}
+
+local function isAtWhitelistedLocation()
```
minipunch, line 32:
```suggestion
local function isAtWhitelistedLocation()
    local playerCoords = GetEntityCoords(PlayerPedId())
```

minipunch (approved):

PR #2042, opened 2023-04-21
[Phone] New Phone System (Again)

maxgestic:


minipunch:
got this when trying to retrieve a vehicle:

minipunch:
> ~got this when trying to retrieve a vehicle:~
> 
> 
> 
> edit: actually it's when u request to get the location of a vehicle listed in the 'out' section which makes more sense

Fixed this

Would be cool to make it so the garage app valet fee is dynamic based on the distance it is from you tho

minipunch:
Also every time I have summoned a vehicle it has just been dropped somewhere nearby without being 'dropped off' by an NPC lol

minipunch:
Seems like the 911 system isn't working though

minipunch:
> Seems like the 911 system isn't working though

fixed

maxgestic:
> Also every time I have summoned a vehicle it has just been dropped somewhere nearby without being 'dropped off' by an NPC lol
> 
> Not a big deal though, seems like maybe it happens if the car was parked at a far away garage or something

Check F8 if the car spawns like that it's usually cos there is an error

minipunch:
getting errors when sending twitter or instagram DMs

going to try using maria db 10 instead of mysql 8 and see if that fixes it

minipunch:
also need to fix how it says "you do not have a phone" every time you join

PR #2053, opened 2023-04-24
update: Chop Shop Improvements / Additions

Savage912-Official:
Changes Made:

- Code Improvements
- Replaced Deprecated Natives
- Replaced __resource.lua with fxmanifest.lua
- Rare Items Table Added (https://usarrp.gg/forums/topic/2150-chop-shop-upgraded-parts/)


resources/usa_chopshop/cl_chopshop.lua
```diff
@@ -23,12 +23,12 @@ Citizen.CreateThread(function()
   local handle = 0
   local display_name = "Undefined"
   while true do
-    Wait(0)
+    Citizen.Wait(0)
```
minipunch, line 26: fyi in the future lets not use the fully qualified name, `Wait` works fine

resources/usa_chopshop/cl_chopshop.lua
```diff
@@ -196,24 +196,25 @@ function addBlips()
...
+  if #(GetEntityCoords(PlayerPedId()) - vector3(x, y, z)) < distance then
+    local onScreen, _x, _y = GetScreenCoordFromWorldCoord(x, y, z)
+    local textLength = string.len(text)
+    local factor = textLength / 430
     SetTextScale(0.35, 0.35)
     SetTextFont(4)
     SetTextProportional(1)
     SetTextColour(255, 255, 255, 215)
-    SetTextEntry("STRING")
+    BeginTextCommandDisplayText("STRING")
```
minipunch, line 213: why this change?

resources/usa_chopshop/sv_chopshop.lua
```diff
@@ -5,12 +5,25 @@ local items = {
   {name = "Iron Oxide", legality = "legal", quantity = 1, type = "misc", weight = 8, price = 950}
 }
 
-RegisterServerEvent("chopshop:reward")
+local rareItems = {
+  { name = "Custom Radio", price = 10000, quantity = 1, weight = 15, type = "mechanicPart", notStackable = true, objectModel = "sm_prop_smug_wall_radio_01" },
```
minipunch, line 9: probably could have used `exports.usa_rp2:getItem(...)` for this

minipunch: Think you can try that and lmk?

minipunch (changes requested):

resources/usa_chopshop/sv_chopshop.lua
minipunch: Actually it would prob be best to store all our items in the DB actually so any script can access them easily

Because now I realized these are defined not in `usa_rp2` so that wouldn't work as I stated, so nvm for now

minipunch (approved):

PR #2058, opened 2023-04-25
Cat Cafe v1.1.0

DatBoiWeep:
- Added check to ensure player does not have a job before working at Cat Cafe (Removes issue when player's get banned for not being in the job)


resources/usa_catcafe/client/main.lua
```diff
@@ -135,16 +135,48 @@ end)
...
-        TriggerServerEvent("catcafe:getrank")
-        TriggerServerEvent('catcafe:retrievestats')
+        local job = TriggerServerCallback {
+            eventName = "catcafe:isPlayerInJob",
+            args = {}
+        }
+        if job then
+            hrNotify("You are currently working for another job. You need to stop working for them first", "error")
+        else
+            TriggerServerEvent('catcafe:checkCriminalHistory')
```
minipunch, line 145: Seems weird that we have 3 separate server events  being triggered here, why not consolidate into a single 'sign on'  event or something

resources/usa_catcafe/server/main.lua
```diff
@@ -447,6 +463,31 @@ AddEventHandler("catcafe:payB", function(bool)
     end
 end)
 
+local firstPlace, secondPlace, thirdPlace, fourthPlace, fifthPlace
+
+RegisterServerEvent("catcafe:updateLeaderboard")
+AddEventHandler("catcafe:updateLeaderboard", function()
```
minipunch, line 469: why a separate event for this? why not just do this in the server callback event?

resources/usa_catcafe/client/main.lua
minipunch: I guess it's OK for now if it's already been like this for this PR but just as an fyi for the future

resources/usa_catcafe/server/main.lua
DatBoiWeep: fixed :salute:

minipunch (approved):

PR #2065, opened 2023-04-27
update: Distance Check Changes

Savage912-Official:
Updated some of the Vdist to use #(vector3 - vector3)

Gonna do more later on just had a little free time so got a few done.


minipunch:
Did you actually go through each script to make sure these changes didn't break anything?

If not then for now let's just close this and not worry about it since I don't think that risk is worth the benefit and I'm not about to go thru them all lol

PR #2069, opened 2023-04-28
Disabled pier script

Zeeroqt:
causing game crashing and it is very known


minipunch:
lots of people don't crash and enjoy the pier stuff (myself included), so for now going to close this

PR #2072, opened 2023-04-29
Updated garage menu

### LINKED ISSUE #1790: filterable vehicle garage list
since it gets pretty unwieldy when dealing with more than a handful of vehicles since u gotta scroll thru them all the time and there is no filter/search

minipunch:
closes #1790

not sortable, but it is scrollable which is a good start

we can add filtering and manual sorting eventually


minipunch:
probably need to manually run `npm install` after pulling

PR #2087, opened 2023-05-03
update: Disable placing self while downed

Savage912-Official:
Disabled placing yourself in a vehicle while downed.

### DIFF
--- a/resources/usa_place/cl_place.lua
+++ b/resources/usa_place/cl_place.lua
@@ -13,27 +13,13 @@ end
 RegisterNetEvent("place:place")
 AddEventHandler("place:place", function(seatOption, placedByLEO, placerID)
 	local me = PlayerPedId()
+	local veh = exports.globals:getClosestVehicle(5.0)
+
 	if DoesEntityExist(me) then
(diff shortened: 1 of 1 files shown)


resources/usa_place/cl_place.lua
```diff
@@ -13,27 +13,13 @@ end
...
+	if DoesEntityExist(me) then
+		if not IsPedDeadOrDying(me, true) then
 			_placeIntoVehicleInternal(me, veh, seatOption)
-			-- re-incapacitate
-			start = GetGameTimer()
-			while GetGameTimer() - start < 1000 do
-				Wait(1)
-			end
-			SetEntityHealth(me, 0)
+		elseif IsPedDeadOrDying(me, true) then
```
minipunch, line 21: P sure this event is called for all places, not just placing yourself (even though the player ped var is named `me`)

I think this makes it so it will never revive the person before placing them? Even when placing others

Savage912-Official: Alright I’ll work on it Monday

PR #2100, opened 2023-05-08
fix: Mechanic Job Veh Spawning

Savage912-Official:
Fixed the mechanic job not spawning the correct vehicles.


resources/usa_mechanicjob/mechanic_cl.lua
```diff
@@ -519,8 +519,8 @@ function SpawnHeavyHauler(coords)
 	end)
 end
 
-function SpawnTowFlatbed(coords)
-	local numberHash = 1353720154 -- tow truck
+function SpawnTowFlatbed(coords, model)
```
minipunch, line 522: I think ideally we combine this function into a more generic one that spawns any kind of truck, but for now this is fine since that is how it was already

resources/usa_mechanicjob/mechanic_cl.lua
```diff
@@ -376,9 +376,9 @@ AddEventHandler("mechanic:spawnTruck", function(repairCount)
 		local dutyCoordsVector = vector3(info.duty.x, info.duty.y, info.duty.z)
 		if Vdist(mycoords, dutyCoordsVector) < 10 then
 			if repairCount >= Config.LEVEL_3_RANK_THRESH then
-				SpawnHeavyHauler(info.truck_spawn)
+				SpawnHeavyHauler(info.truck_spawn, model)
```
minipunch, line 379: Where is `model` defined ?

resources/usa_mechanicjob/mechanic_cl.lua
Savage912-Official: Yeah if you wanna just approve this or however I’ll combine them both tomorrow into one function

resources/usa_mechanicjob/mechanic_cl.lua
Savage912-Official: I don’t think this event is even used is it?

minipunch: Not sure

Savage912-Official: It doesn’t seem like ktnis

Savage912-Official: Looks like it used to be but was replaced by mechanic:openTruckSpawnMenu

resources/usa_mechanicjob/mechanic_cl.lua
```diff
@@ -543,16 +515,18 @@ function SpawnTowFlatbed(coords)
...
 		}
 
 		-- give key to owner
 		TriggerServerEvent("garage:giveKey", vehicle_key)
-		TriggerServerEvent('mdt:addTempVehicle', 'MTL Flatbed', "Bubba's Mechanic Co.", vehPlate)
+		TriggerServerEvent('mdt:addTempVehicle', vehType, "Bubba's Mechanic Co.", vehPlate)
 
 		-- give repair kit to start with --
-		TriggerServerEvent("mechanic:giveRepairKit", vehPlate)
+		if vehType == "TowFlatbed" then
```
minipunch, line 527: Why only give repair kit for the flatbed? Prob want it in all spawned vehicles

minipunch (approved):

minipunch (approved):

PR #2102, opened 2023-05-08
fix: Ability to Rob Animals

Savage912-Official:
Fixed the ability to rob animals.

The function was already in place it was just removed when the ability to take NPC's hostage was added.


minipunch:
Wonder why the check was removed and if it was on purpose for some reason lol...

PR #2112, opened 2023-05-11
Nerf Mining

OffixialMikey:
Nerfing mining to encourage other civilians to try other jobs besides just mining since this is the best paying job.


minipunch:
Haven't seen or heard anyone else say we should nerf mining, going to close this for now

if u think it's too OP then maybe provide some stats like time it takes to make X money

PR #2122, opened 2023-05-17
update: Reset Char Properties/Businesses on Delete

### LINKED ISSUE #2111: properties don't go up for sale if owner char deleted
haven't tested myself but has been reported by some

@maxgestic wanna look since u worked on this most recently?

Savage912-Official:
When a character is deleted their properties and businesses will now go up for sale as they should.

Also changed the character-selection js so when someone is creating a character and they hit backspace it won't go back to the main screen as it just causes and issue.

closes #2111


minipunch:
> "they hit backspace it won't go back to the main screen as it just causes and issue."

What issue are u talking about ?

Savage912-Official:
> > "they hit backspace it won't go back to the main screen as it just causes and issue."
> 
> What issue are u talking about ?

Like if you are typing your character info and misspell something you hit backspace and it takes you back to the Home Screen causing you to have to re-enter you characters info.

Savage912-Official:
Not that big of a deal just something that could be prevented

resources/usa-properties-og/server.lua
```diff
@@ -1037,6 +1037,39 @@ function Evict_Owners()
 end
 -----------------------------------------
 
+RegisterNetEvent("usa-properties-og:server:resetcharproperties")
+AddEventHandler("usa-properties-og:server:resetcharproperties", function(char_identifier)
+  for name, info in pairs(PROPERTIES) do
+    if info.type == "house" then
```
minipunch, line 1043: Why the if statement to check if it is a house or business? Both branches appear to be running the same code?

resources/usa-properties-og/server.lua
```diff
@@ -1037,6 +1037,39 @@ function Evict_Owners()
...
+      if PROPERTIES[name].owner.identifier == char_identifier then
+        PROPERTIES[name].fee.end_date = 0
+        PROPERTIES[name].fee.due_days = 0
+        PROPERTIES[name].fee.paid_time = 0
+        PROPERTIES[name].fee.paid = 0
+        PROPERTIES[name].will_leave = false
```
minipunch, line 1049: I guess this line is the only difference?

Savage912-Official: Yeah u think we just combine both?

minipunch: yeah, just did it

minipunch (approved):

PR #2123, opened 2023-05-17
fix: Selling mined goods in vehicles

Savage912-Official:
Made it so you can no longer sell mined goods while in a vehicle.


resources/usa_mining/client/main.lua
```diff
@@ -83,9 +83,10 @@ end)
 
 Citizen.CreateThread(function()
     while true do
+        local player = PlayerPedId()
```
minipunch, line 86:
```suggestion
```

No need to call this native every tick

resources/usa_mining/client/main.lua
```diff
@@ -83,9 +83,10 @@ end)
...
         if nearMarker(purchase_location.x, purchase_location.y, purchase_location.z, 3) then
             exports.globals:DrawText3D(purchase_location.x, purchase_location.y, purchase_location.z, '[E] - Sell Goods | [H] - Hint')
-            if IsControlJustPressed(0, 86) then
+            if IsControlJustPressed(0, 86) and not IsPedInAnyVehicle(player) then
```
minipunch, line 89:
```suggestion
            if IsControlJustPressed(0, 86) and not IsPedInAnyVehicle(PlayerPedId()) then
```

only call it at minimum when necessary instead (like once they press the hotkey here)

minipunch (approved):

PR #2132, opened 2023-06-01
BMW Shop Update

Urbaniized:
Moved all cars from BMW dealership to Benefactor
Stopped BMW Shop script

#2131

PR #2133, opened 2023-06-01
Z3 Handling Update

Urbaniized:
Changed Z3 to RWD


minipunch (commented):
looks like these changes snuck into #2132

so gonna close this

PR #2136, opened 2023-06-04
Revert to old prison

DatBoiWeep:
Because the community said yes


minipunch:
looks like we forgot to remove some of the old stuff like the b-ball court, wonder if same goes for the TV, phone, peds and what not ?

minipunch (approved):

PR #2184, opened 2023-07-24
QoL Updates

Urbaniized:
Removed some cars from shops to help QoL

Cars can still be obtained from players that own them or through boosting.


minipunch (approved):

minipunch:
syntax error in 3... 2...

PR #2212, opened 2023-10-22
Ensure Spooky Szn

DatBoiWeep:
Spooky, scary skeletons
Send shivers down your spine
Shrieking skulls will shock your soul
Seal your doom tonight
Spooky, scary skeletons
Speak with such a screech
You'll shake and shudder in surprise
When you hear these zombies shriek
We're sorry skeletons, you're so misunderstood
You only want to socialize, but I don't think we should
'Cause spooky, scary skeletons
Shout startling, shrilly screams
They'll sneak from their sarcophagus
And just won't leave you be
Spirits supernatural are shy what's all the fuss?
But bags of bones seem so unsafe, it's semi-serious
Spooky, scary skeletons
Are silly all the same
They'll smile and scrabble slowly by
And drive you so insane
Sticks and stones will break your bones
They seldom let you snooze
Spooky, scary skeletons
Will wake you with a boo!


DatBoiWeep:
Already exists 😭

PR #2214, opened 2023-12-08
Impound Update

Urbaniized:
Updated police impound command to have a wait time of `60 seconds` before the impound is complete and converted to ox stuff.

Created *new* (old) command for `staff` and `eventPlanners` to not have a wait time.


resources/usa_rp2/server.lua
```diff
@@ -274,6 +267,32 @@ AddEventHandler("impound:impoundVehicle", function(vehicle, plate)
...
 
+TriggerEvent('es:addCommand', 'dv', function(source, args, char)
+    local group = exports["essentialmode"]:getPlayerFromId(source).getGroup()
+    if group ~= "user" or char.get("job") == "eventPlanner" then
+        TriggerClientEvent('staffimpoundVehicle', source)
+    end
+end, { help = "(Delete) Impound a vehicle. DO NOT USE FOR TRAINS USE /dt INSTEAD" })
+
+RegisterServerEvent("impound:staffimpoundVehicle")
+AddEventHandler("impound:staffimpoundVehicle", function(vehicle, plate)
```
minipunch, line 278: I don't think we need to create a whole separate event with duplicated code

Why not just use the existing event and tack on `eventPlanner` as a permissible job

resources/usa_rp2/server.lua
```diff
@@ -248,19 +248,12 @@ TriggerEvent('es:addJobCommand', 'impound', { "sheriff", "ems", "corrections" },
...
-	end
-end, { help = "(Delete) Impound a vehicle. DO NOT USE FOR TRAINS USE /dt INSTEAD" })
-
 RegisterServerEvent("impound:impoundVehicle")
 AddEventHandler("impound:impoundVehicle", function(vehicle, plate)
 	local user = exports["essentialmode"]:getPlayerFromId(source)
 	local userGroup = user.getGroup()
 	local playerJob = exports["usa-characters"]:GetCharacterField(source, "job")
-	if playerJob == "mechanic" or playerJob == "sheriff" or playerJob == "ems" or playerJob == "corrections" or userGroup == "owner" or userGroup == "admin" or userGroup == "mod" or userGroup == "superadmin" then
+	if playerJob == "mechanic" or playerJob == "sheriff" or playerJob == "ems" or playerJob == "corrections" then
```
minipunch, line 256: Just keep the admin groups and add the `eventPlanner` job  here so u can use this event

resources/usa_rp2/client.lua
```diff
@@ -780,66 +780,181 @@ local distanceToCheck = 5.0
...
+  else
+      exports.globals:notify("No nearby vehicle to impound.", "^3INFO: ^0No nearby vehicle.")
+  end
+end)
+
+-- Register a network event
+RegisterNetEvent( 'staffimpoundVehicle' )
+
+--Staff Impound
+AddEventHandler( 'staffimpoundVehicle', function()
```
minipunch, line 898: Instead of duplicating this code we should just have it skip the progress bars if the person who triggered it is staff

If it's done with a server sided command we can just pass their staff status in as a parameter, otherwise if it is triggered from some client action then we'd probably just use `TriggerServerCallback` to fetch their admin status

minipunch (changes requested):

PR #2221, opened 2023-12-13
Update server.cfg

Zeeroqt:
removed mlo's that sit with no purpose atm


server.cfg
```diff
@@ -94,7 +94,7 @@ start map_parsons
 #start josef_helipad
 #start crackhouse
 #start smokey_grove
-start secretbiker
+#start secretbiker
```
minipunch, line 97: Is this the one in Sandy Shores where the weapons extras dealer is ?

Zeeroqt: yessir

minipunch (approved):

PR #2224, opened 2023-12-18
Update Gun Shop

Zeeroqt:
new gun store update for usarrp


Zeeroqt:
https://imgur.com/guVedjl

minipunch:
Hm interesting approach, using the general store script

but 1) it's missing a couple server sided actions like registering the gun in the `legalweapons` DB (which the MDT gun search uses) for example (see `gunShop:requestPurchase`)

and 2) it's missing the valid gun permit check

if you can port over those checks then perhaps this could work

minipunch (changes requested):
see comment

PR #2229, opened 2023-12-20
Mining Update

OffixialMikey:
Try to boost the Players to bring back Grinding for money instead of only doing Art heist, Drugs, And the new jobs in the server.

### DIFF
--- a/resources/usa_mining/server/server.lua
+++ b/resources/usa_mining/server/server.lua
@@ -1,16 +1,16 @@
 math.randomseed(os.time())
 
 local common = {
-    {name = "Bauxite", type = "misc", price = 175, legality = "legal", quantity = 1, weight = 5.0},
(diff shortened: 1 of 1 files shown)


Zeeroqt:
its perfect how it is. you dont see from a stand point on how when servers are really grindy people wont play because they waste their time to get x amount. all these pr's are number changes

OffixialMikey:
> its perfect how it is. you dont see from a stand point on how when servers are really grindy people wont play because they waste their time to get x amount. all these pr's are number changes

They clearly need to be brought back. People wont do it because the prices aren't good. No need for you to respond on my stuff

Zeeroqt:
its the whole economy that needs to be reset with everything changed but its mining is quite perfect how it is right now. once ecom is wiped and everything is changed thats when the role of jobs need to be changed

minipunch (approved):

PR #2232, opened 2023-12-20
increase container pay

Crownapple:
I believe should increase due to new job in the city that pay double if not triple this job


Zeeroqt:
400 sounds reasonable 500 is pushing it

Crownapple:
> 400 sounds reasonable 500 is pushing it

I think you need to read the whole thing before you start commenting on anything.

minipunch:
> new job in the city that pay double if not triple this job

What job u referring to ?

Crownapple:
> > new job in the city that pay double if not triple this job
> 
> What job u referring to ?

construction you make about $900^ pre point and it takes about 5 seconds or so and for doc job/ container you make about anyway for 800 to 1.3k and takes around anyway from 35 secs to a min or so depending on the where the container is located

resources/ContainerForklift/server.lua
```diff
@@ -94,10 +94,10 @@ AddEventHandler("containerjob:reward", function(securityToken)
         targetContainer = onJob[source]
         -- calculate reward based on distance from drop off
         local dist = #(vector3(targetContainer.x, targetContainer.y, targetContainer.z) - vector3(CONTAINER_DROP_OFF.x, CONTAINER_DROP_OFF.y, CONTAINER_DROP_OFF.z))
-        local reward = math.floor(dist * 3.5)
+        local reward = math.floor(dist * 4.5)
```
minipunch, line 97:
```suggestion
        local reward = math.floor(dist * 4.15)
```

resources/ContainerForklift/server.lua
```diff
@@ -94,10 +94,10 @@ AddEventHandler("containerjob:reward", function(securityToken)
...
         -- give reward
         local char = exports["usa-characters"]:GetCharacter(source)
-        char.giveBank(reward + math.random(0, 250), "Container Reward")
+        char.giveBank(reward + math.random(0, 500), "Container Reward")
```
minipunch, line 100:
```suggestion
        reward = reward + math.random(0, 500)
        char.giveBank(reward, "Container Reward")
```

minipunch (approved):

PR #2236, opened 2023-12-27
Increase meth payout

Crownapple:


minipunch:
It was recommended that maybe instead of increasing the payout we could/should reduce the explosion chance instead since some are saying it still explodes too frequently

Crownapple:
> It was recommended that maybe instead of increasing the payout we could/should reduce the explosion chance instead since some are saying it still explodes too frequently

Maybe make a vote on it to see what people want

PR #2240, opened 2023-12-29
Gun Damage Update

Crownapple:


GordoHockey4:
I like the idea of lowering the Goldbeast damage slightly, but I do not think the Carbine mk2 need increasing....unfair advantage for PD

Hyde0611:
> I like the idea of lowering the Goldbeast damage slightly, but I do not think the Carbine mk2 need increasing....unfair advantage for PD

Reallistically, PD would have an advantage IRL. The goal is to try to avoid shootouts. I do think that Carbine should be lowered because crims have access to those as well. Essentially make it to where the Carbine MK2 has a slightly higher multiplier. Maybe have it increased by .2 or .3 because the amount of crims that are actively running around shooting cops and other people with Carbines and Goldbeasts is ridiculous.

Crownapple:
> I like the idea of lowering the Goldbeast damage slightly, but I do not think the Carbine mk2 need increasing....unfair advantage for PD

If you dont know what the code means/what is does I would like you to reframe on commenting on my pull request. there is other things that play a factor on a gun more then damage modifiers

Khajit71:
goldbeast is hard af to craft, especially the gold for it, takes A LONG time. To follow up on what Gordon said, either make shit fair across the board or don't, I haven't personally read the code myself, it's what I always go by however.

Hyde0611:
Maybe lower the Akorus slightly as well, to look something like this, granted it's difficult to make, it still hits like a truck, because we know damn well if we lower the Goldbeast people are gonna run around with either a Carbine or an Akorus. Would make it something like this just to not eliminate shootouts but make it a tad more fair on all ends.

[`WEAPON_CARBINERIFLE`] = 1.5
[`WEAPON_CARBINERIFLE_MK2`] = 1.7
[`WEAPON_M4GOLDBEAST`] = 1.5
[`WEAPON_AKORUS`] = 1.5

Reallistically I think that's quite fair. Their guns will still hit decently and some of PD will have access to a weapon that hits slightly harder because I don't think that a .2 increase in the mutiplier will make that much of a difference.

Hyde0611:
Considering the latest change in crims no longer being able to purchase gun parts, that actually makes the Akorus and Goldbeasts easier to get compared to a Carbine

Khajit71:
> Considering the latest change in crims no longer being able to purchase gun parts, that actually makes the Akorus and Goldbeasts easier to get compared to a Carbine

yea but they ain't got access to mk2 carbine

Khajit71:
but pd has access to sniper :)

GordoHockey4:
I am fine with lowering the goldbeast slightly...still don't think PD guns should get increased

PR #2255, opened 2024-01-07
Hospital bed changes

DatBoiWeep:
-  Removes players who are hand cuffed from the bed AFTER treatment has been completed. This prevents players from AFKing to avoid police interactions.


minipunch (approved):

minipunch:
Nice

PR #2262, opened 2024-01-16
Lb phone update 1.4.5

minipunch:
spurred on by videos seeming not to work

it worked fine after this update on my test server, so we'll see if it fixes it


minipunch:
just need to merge this and then run all the .sql update files

PR #2290, opened 2024-02-28
evidence menu added inside new Davis SO

kian969:
don't judge if this don't work yo... i just pretend to know what i'm doing :)


minipunch (approved):
nice!!!

PR #2301, opened 2024-03-03
Weapon Damage Update

Urbaniized:
Updated weapon damage for SMGs & ARs

Tested all the SMGs & ARs, reduced damage across the board to prolong gun fights.

Average time to kill before: .50 - .85 seconds
Average time to kill now: 1.50-1.85 seconds

Average time to kill with a pistol: 1.95 - 2.50 seconds (untouched)

Damage and time to kill will vary depending on the distance to the target.

This is just to start and can re-adjust as needed down the road.


resources/usa_rp2/client/cl_weaponDamageModifier.lua
```diff
@@ -66,8 +66,8 @@ local WEAPON_DAMAGE_MODIFIERS = {
   [`WEAPON_POOLCUE`] = 0.1,
   [`WEAPON_KATANAS`] = 0.5,
   [`WEAPON_SHIV`] = 0.7,
-  [`WEAPON_M4GOLDBEAST`] = 1.5,
-  [`WEAPON_AKORUS`] = 1.7,
+  [`WEAPON_M4GOLDBEAST`] = 0.45,
```
minipunch, line 69: basically just a gold carbine right? why not make it the same damage?

Urbaniized: > basically just a gold carbine right? why not make it the same damage?

Recoil on it is next to none allowing people to spray in a tight circle even at a distance.

minipunch (approved):

PR #2304, opened 2024-03-04
Doormanager copy door info to clipboard command

kian969:
I really don't know if this will work but it's the only thing that makes sense to me - with all these new MLO's being added and removed frequently it would be grand to have the door manager "copy door info to clipboard" feature at admin level instead of owner (p.s. i can't be bothered to make my own server/copy the repo across to a server, it's a lot of effort) 🤣

### DIFF
--- a/resources/usa_doormanager/sv_doormanager.lua
+++ b/resources/usa_doormanager/sv_doormanager.lua
@@ -662,13 +662,13 @@ AddEventHandler("doormanager:firstJoin", function()
   TriggerClientEvent("doormanager:update", source, DOORS)
 end)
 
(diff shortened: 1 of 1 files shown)


DatBoiWeep (commented):
It'll work. But, is the purpose of this only to allow Admins+ to copy the code to edit the usa_doormanager config?

kian969:
> It'll work. But, is the purpose of this only to allow Admins+ to copy the code to edit the usa_doormanager config?

Well I just thought I can’t really think of anyone below Admin that would do it so I said sod it I’ll just do it😂

minipunch (approved):

PR #2325, opened 2024-03-08
Add clothing store to auto exotic and cayo perico island

kian969:


minipunch (approved):
I only see an addition for cayo perico fyi @kian969 but I'll approve this for now at least

Issue #2323, opened 2024-03-07
limit robbery reward when no cops on


minipunch:
rn it gives the same amount regardless of any cops online or not

should give a bit less if no cops online

GordoHockey4:
If I change the policeNeededForBonus = 3    ---->   policeNeededForBonus = 0     AND THEN changed  bonus = math.floor(reward * 0.40)   ---->  to have a negative 0.40 or something of the sorts...would that work???

GordoHockey4:
OR would that just make anytime there is 0 to however many cops online...a deduction in robbery earnings

GordoHockey4:
Attempted to add a bonus if cops are online...we can reduce values if needed to make it lower when cops offline...again all just an attempt...I am very new at this lol and have a LOT to learn

minipunch:
[716fb1d](https://github.com/minipunch/usarealismrp-fxserver/commit/716fb1de7ee3e8cd2c732f9d28853019465183e5)

PR #2327, opened 2024-03-09
Fixes Player Owned Business Robbery Bonus When Cops Online  | (helps with #2323)

GordoHockey4:
usa-businesses | changed:

policeNeededforBonus = 0
POLICE_NEEDED = 0

(CL and SV config)

usa-businesses/robberies | changed:

bonus = math.floor(reward * 0.50)

**in theory now when there are more than 0 cops online, there should be a bonus applied to robberies**

**also cops can no longer start a store robbery by aiming their gun when breaching a store**

### DIFF
--- a/resources/usa-businesses/robberies/sv_robbery.lua
+++ b/resources/usa-businesses/robberies/sv_robbery.lua
@@ -52,7 +52,7 @@ AddEventHandler('business:finishRobbery', function(storeName)
 				local policeOnline = exports["usa-characters"]:GetNumCharactersWithJob("sheriff")
 				local bonus = 0
 				if policeOnline >= policeNeededForBonus then
-					bonus = math.floor(reward * 0.40)
+					bonus = math.floor(reward * 0.50)
 				end
 				char.giveMoney(reward + bonus)
 				print("ROBBERY: "..GetPlayerName(usource)..'['..GetPlayerIdentifier(usource).."] has been rewarded reward: "..(reward + bonus))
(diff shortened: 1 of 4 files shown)


DatBoiWeep:
I am confused as to why you are setting the number of police needed for robbing stores.

The purpose of store robberies bonuses is to give criminals an advantage for robbing a store with cops on.

resources/usa-businesses/robberies/sv_robbery.lua
```diff
@@ -52,7 +52,7 @@ AddEventHandler('business:finishRobbery', function(storeName)
 				local policeOnline = exports["usa-characters"]:GetNumCharactersWithJob("sheriff")
 				local bonus = 0
 				if policeOnline >= policeNeededForBonus then
-					bonus = math.floor(reward * 0.40)
+					bonus = math.floor(reward * 0.50)
```
DatBoiWeep, line 55: This is increasing the bonus, which is already 100% chance per the edits you made in `sv_config.lua` > line 2

DatBoiWeep: Meaning, no cops are needed for a bonus to apply

DatBoiWeep:
Also, I don't think that's what is being requested in #2323 because it looks like Mini wants it to scale per the number of cops on and lower itself when no cops are on.

GordoHockey4:
CLOSING THIS BECAUSE I HAVE A SMALL BRAIN AND ITS ALL WRONG

PR #2335, opened 2024-03-12
fix broken item images pt3

kian969:
Didn't fix the cards because that's a whole lot of effort and the majority of the cards are for players that aren't even here anymore...

just need to add the files to the **images** folder


kian969:
@minipunch it's giving me "Uploads are disabled" when I try to upload all of the images I just reset in this commit 😢

minipunch (approved):

PR #2347, opened 2024-03-16
New Bahama Mama

Urbaniized:
Disabled Gabz Bahama and Enabled new mlo


minipunch:
What's with all those changes from @kian969 on the `custom_maps` repo?

A bunch of deleted files and what not

Are we sure we know those work and aren't by accident?

kian969:
IMG_2187

If you’re talking about these, when I did the “Upload files” thing I did it in the main repo instead of inside the Auto Exotics file so I went through and deleted the texture files I uploaded and then reuploaded them inside of the Auto Exotics file, I didn’t delete anything that was already there it was just an accident 👍

